PR #13, opened 2017-10-17
updated sass-lint and fixed most problems

yodasw16:
Still needed:

- [x] Fix regex to account for responsive suffixes
- [x] After Scales updates, need to update HorizontalList mixin in pagination so it uses namespaces


yodasw16:
String interpolation doesn't work on mixins so using namespaces ins't possible

yodasw16:
Before we merge this can one of you pull down this branch and make sure the thing runs? I had a weird conflict and I'm not sure I didn't break things. After this is merged I will create a new release. @afebbraro @RCopeland

packages/spark-extras/components/description-table/_description-table.scss
```diff
@@ -1,48 +1,55 @@
-// Variables
+// ==================================================================
+// Date Picker
```
RCopeland: Should be Description Table

packages/spark-extras/components/highlight-board/_highlight-board.scss
```diff
@@ -1,4 +1,10 @@
-// Variables
+// ==================================================================
+// Navigation
```
RCopeland: Should be Highlight Board

PR #20, opened 2018-01-05
updated marked to 0.3.6, per gh recommendation

RCopeland:


package.json
```diff
@@ -51,5 +51,8 @@
     "start": "gulp --dev",
     "build": "npm run prestart && gulp",
     "docs": "doctoc --title \"### Contents\" --maxlevel 2 docs/README.md"
+  },
+  "dependencies": {
+    "marked": "^0.3.6"
```
yodasw16: I just noticed that the recommendation is to update to 0.3.9, not 0.3.6 which is the version with the vulnerability.

PR #27, opened 2018-02-02
added purple dark and created setup-core file

RCopeland:


packages/spark-core/settings/_colors.scss
```diff
@@ -11,6 +11,7 @@ $blue-medium: rgb(32, 129, 191);
 $blue-dark: rgb(12, 69, 105);
 
 $purple: rgb(85, 19, 135);
```
yodasw16: I would also change this to `$purple-medium` to match the convention.

packages/spark-core/setup-core.scss
```diff
@@ -0,0 +1,62 @@
+///
```
yodasw16: Should we have 3 files instead of 1? Maybe `_scales.scss`, `_spark-core.scss`, and `spark.scss`.

Then the main import file only needs a couple lines? Is that more or less confusing?

packages/spark-core/setup-core.scss
```diff
@@ -0,0 +1,62 @@
...
+// Generic utility styles
+///
+
+@import "../@scalescss/generic-normalize/normalize";
+@import "../@scalescss/generic-reset/reset";
+
+///
+// Spark-core settings
+///
+@import './settings/colors';
```
yodasw16: I'd move this and settings above everything else since they don't output any code themselves. Also, reset has options that can be replaced by settings.

packages/spark-core/setup-core.scss
```diff
@@ -0,0 +1,62 @@
...
+@import './components/modals';
+@import './components/navigation';
+@import './components/notifications';
+@import './components/pagination';
+@import './components/spinners';
+@import './components/tables';
+@import './components/tabs';
+@import './components/validation-criteria';
+
+@import './tools/animations';
```
yodasw16: This should move up before base stuff. If not, maybe animations aren't tools? Currently it's a function that doesn't output any code itself, so fits in to tools.

PR #75, opened 2018-02-16
added a drizzle-list pattern, updated docs for type

RCopeland:


packages/spark-core/tools/_mixins.scss
```diff
@@ -18,3 +18,43 @@
     margin-left: auto;
   }
 }
+
```
yodasw16: I know this is nit-picky, but can you add a `//` here for consistency?

RCopeland: just added

PR #82, opened 2018-02-20
Dictionary

RCopeland:


src/templates/collection.hbs
```diff
@@ -12,11 +12,11 @@
           {{/each}}
         </ul>
     {{/if}}
-    {{#if nonos}}
-        <h3 class="drizzle-h3">No-no's</h3>
-        <ul>
-          {{#each nonos}}
-            <li>{{.}}</li>
+    {{#if restrictions}}
```
yodasw16: lol. you don't like no-no's?

src/patterns/components/dictionary/base.hbs
```diff
@@ -0,0 +1,33 @@
+---
+order: 1
+notes: This is the default use of this pattern.
+---
+
+<div class="sprk-c-Dictionary">
+  <dl class="sprk-c-Dictionary__keyvaluepairs">
+    <div class="sprk-c-Dictionary__keyvaluepair">
+      <dt class="sprk-c-Dictionary__key sprk-b-Label">Email Address</dt>
+      <dd class="sprk-c-Dictionary__value">bobwise@quickenloans.com</dd>
```
yodasw16: Let's change this email to the spark email or something

src/patterns/components/dictionary/base.hbs
```diff
@@ -0,0 +1,33 @@
+---
```
yodasw16: I think we should standardize what we call the default pattern in a collection. I think I usually am more verbose and I have been using "basic". So for this one it would be "Basic Dictionary" and the next "Striped Dictionary", etc.

src/patterns/components/dictionary/collection.yaml
```diff
@@ -0,0 +1,9 @@
+notes: |
+  This spark-extras pattern is intended for displays of key-value pairs.
```
yodasw16: We need something to visually separate things like package names so they stand out form the rest of the paragraph. There is the class `drizzle-CodeHighlight` that puts a grey background behind it, maybe that?

yodasw16: Also, we should avoid using "pattern". Most of the time "component" should be used instead.

src/patterns/components/dictionary/collection.yaml
```diff
@@ -0,0 +1,9 @@
...
+restrictions:
+   - Each key should have one, and only one, value.
```
yodasw16: "Each key must only have _one_ value"

src/patterns/components/dictionary/base.hbs
yodasw16: Did we remove the heading from this component?

src/patterns/components/dictionary/striped.hbs
```diff
@@ -0,0 +1,34 @@
+---
+order: 2
+notes: This variation is intended for cases where the list of key-value pairs is especially long
```
yodasw16: I don't know what you mean by "especially long". Like lots of entries in the list or the lines are long because it's in a wide container?

yodasw16: If it's the former, I would write: "This variation is for cases where the number of items in the Dictionary exceeds [some number]. It provides a visual cue connecting the key to its value."

If the later, I would write: "This variation is for cases where the width of the Dictionary makes it difficult scan. It provides a visual cue connecting the key to its value."

src/templates/collection.hbs
afebbraro: lol

src/patterns/components/dictionary/base.hbs
afebbraro: Good call. 😂

packages/spark-extras/components/dictionary/_dictionary.scss
```diff
@@ -0,0 +1,44 @@
+// ==================================================================
+// Dictionary
+// ==================================================================
+
+//
+// Available Variables
+///
+$dictionary-border: 1px solid $gray-medium;
```
afebbraro: I like having all the available settings in the settings.scss file since it's easier to make design changes. How do you feel about making a new section under the other component sections in settings for the Dictionary settings?

packages/spark-extras/components/dictionary/_dictionary.scss
```diff
@@ -0,0 +1,44 @@
...
+  @media all and (min-width: $dictionary-breakpoint) {
+    display: table;
+    display: flex;
+    align-items: center;
+  }
+}
+
+// prettier-ignore
+.sprk-c-Dictionary--striped .sprk-c-Dictionary__keyvaluepair:nth-child(even) {
+  background: $dictionary-stripe-color;
```
afebbraro: This comment might deserve the pedantic award haha so I wasn't going to even mention it but I thought maybe it would be a good discussion to bring up as we start to build stuff out. When we are declaring properties that have a shorthand option but we are only using one specific value - do we want to standardize on using the property that matches that value in that case? For this, we could use `background-color: ` instead of the shorthand `background`.  This is something that we could agree on as a team so that our code is consistent going forward. Again, this isn't the hugest deal and not something directly related to this specific PR. @yodasw16  - interested in your thoughts as well.

src/patterns/components/dictionary/collection.yaml
afebbraro: @yodasw16  Rob and I talked yesterday about having a badge or label for the package names so users can see quickly at a glance what package things are for.

packages/spark-extras/components/dictionary/_dictionary.scss
RCopeland: i'll open an issue to talk through this

packages/spark-extras/components/dictionary/_dictionary.scss
yodasw16: I actually think this is a kind of big deal. I agree, we should be using background-color here. Shorthands like background can be problematic if someone writes css that comes later, it might not override it the way they expect.

PR #86, opened 2018-02-21
reduced the banners down to one, updated docs, tests

RCopeland:


packages/spark-core/components/_dividers.scss
```diff
@@ -1,12 +1,9 @@
-.sprk-b-HeadingDivider {
-  border-bottom: 2px solid $gray-medium;
-  margin-bottom: 1rem;
-}
+$divider-border: 2px solid $gray-medium;
 
-.sprk-b-BodyDivider {
+hr,
```
afebbraro: Are we selecting hr's directly?

yodasw16: I think this is another reasonable deviation. Especially since `hr` is so specific and has no side effects if it's styled directly. I'd like to be able to just drop in a totally semantic `hr` and have it look correct.

afebbraro: I dig it!

PR #89, opened 2018-02-23
Icons system, actual icons coming soon

RCopeland:
moved drizzle images out of icons folder, adjusted icons build to just produce icons.svg, manually put icons into default.hbs


src/templates/default.hbs
```diff
@@ -16,6 +16,10 @@
...
...2" id="exclamation" xmlns="http://www.w3.org/2000/svg"><path d="M1088 1248v224q0 26-19 45t-45 19H768q-26 0-45-19t-19-45v-224q0-26 19-45t45-19h256q26 0 45 19t19 45zm30-1056l-28 768q-1 26-20.5 45t-45.5 19H768q-26 0-45.5-19T702 960l-28-768q-1-26 17.5-45t44.5-19h320q26 0 44.5 19t17.5 45z"/></symbol><symbol viewBox="0 0 1792 1792" id="question" xmlns="http://www.w3.org/2000/svg"><path d="M1088 1256v240q0 16-12 28t-28 12H808q-16 0-28-12t-12-28v-240q0-16 12-28t28-12h240q16 0 28 12t12 28zm316-600q0 54-15.5 101t-35 76.5-55 59.5-57.5 43.5-61 35.5q-41 23-68.5 65t-27.5 67q0 17-12 32.5t-28 15.5H804q-15 0-25.5-18.5T768 1096v-45q0-83 65-156.5T976 786q59-27 84-56t25-76q0-42-46.5-74T931 548q-65 0-108 29-35 25-107 115-13 16-31 16-12 0-25-8L496 575q-13-10-15.5-25t5.5-28q160-266 464-266 80 0 161 31t146 83 106 127.5 41 158.5z"/></symbol><symbol viewBox="0 0 1792 1792" id="times" xmlns="http://www.w3.org/2000/svg"><path d="M1490 1322q0 40-28 68l-136 136q-28 28-68 28t-68-28l-294-294-294 294q-28 28-68 28t-68-28l-136-136q-28-28-28-68t28-68l294-294-294-294q-28-28-28-68t28-68l136-136q28-28 68-28t68 28l294 294 294-294q28-28 68-28t68 28l136 136q28 28 28 68t-28 68l-294 294 294 294q28 28 28 68z"/></symbol></svg>
```
RCopeland: I'm not sure how we can get it to import the svg file here. There's a handlebars helper for {{ import }} which would work, but the helpers are handled in a drizzle dep (drizzle-builder)

afebbraro: I took a look last night and came up with adding an icons.hbs partial in the templates dir alongside item.hbs and all that. Then, chaining another pipe command alongside the other icon sprite code in he Gulpfile that pipes it to that icons.hbs file. Only, it doesn’t do a replace contents of icons.hbs. It adds a new file of icons.svg which isn’t a hbs partial so I wasn’t able to import it into the layout. But, I think that’s a good place to start. Have the gulpfile also add the icons.svg in our drizzle app’s templates and import that into our layout.

afebbraro: There’s also {{> filepath/filename}} which should be able to import the file into layout but I wasn’t having luck before I had to log off for the night.

PR #109, opened 2018-03-08
Forms

RCopeland:


packages/spark-core/base/_inputs.scss
```diff
@@ -0,0 +1,105 @@
...
+  }
+}
+
+.sprk-b-TextInput[readonly] {
+  color: $white;
+
+  &:focus {
+    border-color: $gray-medium;
+    box-shadow: 0 0 0 3px $gray-medium;
+  }
```
afebbraro: scales sets border-color: $input-readonly-border-color; and color: $input-readonly-text-color; for this that you could set in our settings. see line 154 of scales forms.scss

packages/spark-core/base/_inputs.scss
```diff
@@ -0,0 +1,105 @@
...
+
+  font-size: $font-size-display-seven;
+  color: $color-display-seven;
+  font-weight: $font-weight-display-seven;
+  line-height: $line-height-display-seven;
+}
+
+.sprk-b-Label--inline {
+  display: inline-block;
+}
```
afebbraro: scales .FormInline sets labels to be inline-block so we don't need this

packages/spark-core/base/_inputs.scss
```diff
@@ -0,0 +1,105 @@
...
+
+.sprk-b-SubLabel {
+  font-size: $font-size-body-two;
+  font-weight: $font-weight-body-two;
```
afebbraro: scales has .FormLabel--additional which, sets the same font-weight you're applying of "normal", that font-weight declaration can be removed

packages/spark-core/base/_inputs.scss
```diff
@@ -0,0 +1,105 @@
...
+
+.sprk-b-TextInput,
+textarea {
+  color: $gray-dark;
+  outline: none;
+  font-size: $font-size-body-one;
+
+  &:focus {
+    border-color: $blue-medium;
+    box-shadow: 0 0 0 3px $blue-medium;
```
afebbraro: setting the box-shadow value to a var might be helpful for theming

packages/spark-core/base/_inputs.scss
```diff
@@ -0,0 +1,105 @@
...
+    position: absolute;
+    top: 0.6rem;
+    left: 0.55rem;
+  }
+}
+
+.sprk-b-TextInput--date-picker {
+  background-image: url('../images/calendar.svg');
+  background-position: 5px 3px;
+  padding-left: 3rem;
```
afebbraro: How does $space-l look for the padding left on this?

packages/spark-core/base/_inputs.scss
```diff
@@ -0,0 +1,105 @@
...
+  display: inline-block;
+}
+
+.sprk-b-SubLabel {
+  font-size: $font-size-body-two;
+  font-weight: $font-weight-body-two;
+}
+
+.sprk-b-TextInput,
+textarea {
```
afebbraro: You'll want to add `select` here too.

packages/spark-core/base/_inputs.scss
```diff
@@ -0,0 +1,105 @@
...
+  margin-bottom: $space-m;
+}
+
+.sprk-b-HelperText {
+  color: $gray-medium;
+}
+
+.sprk-b-ErrorText {
+  color: $red-medium;
+  font-size: $font-size-body-two;
```
afebbraro: These should inherit the base-font-size, line height, weight from the `body` (scales settings) right? So, I don't think we need these declarations but I would double check.

PR #112, opened 2018-03-09
Forms

RCopeland:


packages/spark-core/base/ssnInput.js
```diff
@@ -0,0 +1,45 @@
...
+    });
+  });
+}
+
+function formatSSN (element) {
+  let val = element.value;
+  let newVal = '';
+  const sizes = [3, 2, 4];
+
+  if (!isNaN(val) && val.length === 9) {
```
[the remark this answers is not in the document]
afebbraro: isNaN() has some interesting results like it will return false in the case of a null value `isNaN(null)` or empty string `isNaN('')`.  Which, we wouldn't want to happen. But, since you're also checking the length that would rule out those two scenarios. But, back to the "interesting results" bit that I mentioned. This will also return false `isNaN('0xF111111')`. isNaN thinks that '0xF111111' is a number. Which, results in those 9 chars passing your if statement isNaN and length check and getting formatted. We might want to add an enhanced check when we do get to breaking this out to a validator. 
screen shot 2018-03-13 at 9 11 00 pm

src/patterns/components/inputs/radios-disabled.hbs
```diff
@@ -0,0 +1,27 @@
+---
+name: Radio Buttons - Disabled State
+notes: |
+  These are radio buttons. They use the native browser styles.
```
afebbraro: Might be redundant to restate "These are radio buttons".

src/patterns/components/inputs/radios-error.hbs
```diff
@@ -0,0 +1,28 @@
+---
+name: Radio Buttons - Error State
+notes: |
+  These are radio buttons. They use the native browser styles.
```
afebbraro: Might be redundant to restate "These are radio buttons".

src/patterns/components/inputs/radios.hbs
```diff
@@ -2,26 +2,25 @@
 name: Radio Buttons
 notes: |
   These are radio buttons. They use the native browser styles.
```
afebbraro: Might be redundant to restate "These are radio buttons".

src/patterns/components/inputs/select-disabled.hbs
```diff
@@ -0,0 +1,21 @@
+---
+name: Select Box - Disabled State
+notes: |
+  This is a select box.
```
afebbraro: We could suggest some examples where they might want to use a select box here.

src/patterns/components/inputs/select-error.hbs
```diff
@@ -0,0 +1,22 @@
+---
+name: Select Box - Error State
+notes: |
+  This is a select box.
```
afebbraro: Might be redundant to restate "This is a select box".

src/patterns/components/inputs/ssn-disabled.hbs
```diff
@@ -0,0 +1,22 @@
+---
+name: SSN Input - Disabled State
+notes: |
+  This is an SSN input.
```
afebbraro: Might be redundant to restate "This is an SSN input.".

src/patterns/components/inputs/ssn-disabled.hbs
```diff
@@ -0,0 +1,22 @@
...
+  - On blur of the text input, if the "Show SSN" box is checked, add dashes between the 3rd and 4th number and the 5th and 6th number (ex. 000-00-0000).
+  - 'Placeholder text should read: 000-00-0000.'
+  - On key press, reveal the placeholder inline with the label.
+invisible: true
+order: 11
+---
+
+<div class="sprk-b-InputContainer" data-sprk-input="ssn">
+  <label for="ssn">Social Security #</label>
+  <input disabled class="sprk-b-TextInput sprk-u-Size-100 js-ssn" id="ssn" type="text" placeholder="000-00-0000">
```
afebbraro: You forgot to remove `js-ssn` from the class list.

src/patterns/components/inputs/ssn-disabled.hbs
```diff
@@ -0,0 +1,22 @@
...
+  <div class="sprk-b-SelectionContainer">
+    <input disabled id="show-ssn" type="checkbox" class="js-remove-mask">
```
afebbraro: You forgot to remove  js-remove-mask from the class list.

src/patterns/components/inputs/ssn-error.hbs
```diff
@@ -0,0 +1,24 @@
+---
+name: SSN Input - Error State
+notes: |
+  This is an SSN input.
```
afebbraro: Might be redundant to restate "This is an SSN input.".

src/patterns/components/inputs/textarea.hbs
```diff
@@ -3,11 +3,9 @@ name: Textarea
 notes:
   This is a textarea.
```
afebbraro: We could give some suggestions on when to use a textarea over the others here.

src/templates/default.hbs
```diff
@@ -23,6 +23,5 @@
   {{/content}}
   {{#content "foot"}}
     <script src="{{@root.baseurl}}/assets/drizzle/scripts/drizzle.js"></script>
-    <script src="{{@root.baseurl}}/assets/toolkit/scripts/toolkit.js"></script>
```
afebbraro: nice catch!

PR #127, opened 2018-03-21
Required validation - text inputs / radios / checkboxes, fixed datepicker issues

RCopeland:


packages/spark-core/components/pagination.js
```diff
@@ -35,23 +35,27 @@ export default () => {
...
-    // Initially set aria labels on page load
-    setAriaLabel(defaultPagItem);
-  });
-
-  longPagItems.forEach((longPagItem) => {
-    setAriaLabel(longPagItem);
-    longPagItem.addEventListener('click', () => {
-      setAriaLabel(longPagItem);
+  if (defaultPagItems) {
+    defaultPagItems.forEach((defaultPagItem) => {
```
afebbraro: Are the changes in this file because you forgot to rebase your forms stuff and merged stuff into staging or is it some weird git history thing lol?

src/pages/tests/inputs.hbs
```diff
@@ -3,16 +3,58 @@ title: Inputs Tests
 layout: blank
 ---
 <h1 class="drizzle-h1"> Inputs </h1>
+<form action="" novalidate>
+    {{> drizzle.spark-core-icons}}
```
afebbraro: We should add icons to the tests layout.

src/patterns/components/inputs/collection.yaml
```diff
@@ -13,10 +13,14 @@ contents:
   Money Input: inputs.html#monetary-input
   Password Input: inputs.html#password-input
   Helper Text: inputs.html#helper-text
-  Phone Number: inputs.html#phone-input
+  Phone Number: inputs.html#phone-number-input
   Date Input: inputs.html#date-input
   Date Picker: inputs.html#date-picker
 restrictions:
+  - You should add the 'novalidate' attribute to your form, since we handle form validation ourselves,
```
afebbraro: Glad you added this to restrictions! Should help apps understand more.

packages/spark-core/base/requiredTextInput.js
```diff
@@ -0,0 +1,34 @@
+import getElements from '../utilities/getElements';
+import setValidTextInput from '../utilities/validation/setValidTextInput';
+import setInvalidTextInput from '../utilities/validation/setInvalidTextInput';
+
```
afebbraro: This all looks good. We should in the future add some comment blocks describing what the code is for.

packages/spark-core/components/pagination.js
afebbraro: Oh, I see that you added in a check for if the elems exist. Good catch. Though, I think the check should be done on longPaginations and defaultPaginations. Because if the main pagination elements don't exist then we don't need to check if the items inside exist.

RCopeland: i just added that cause i was getting forEach is not a function on pages other than the pagination page

PR #131, opened 2018-03-22
consolidated tab templates, added js for switching

RCopeland:


packages/spark-core/components/_tabs.scss
```diff
@@ -67,6 +67,7 @@
   color: $white;
 }
```
afebbraro: Could use the spacing vars in this file for the margins/padding.

packages/spark-core/components/tabs.js
```diff
@@ -0,0 +1,117 @@
+/* global window */
```
afebbraro: I don't think we settled on what we were gonna do about doc blocks. I think we should have them - what do you think?

PR #133, opened 2018-03-27
Add Toggle Component

afebbraro:
- add test page for toggle
- add icons and drizzle scripts to load on test pages(blank layout) and remove from regular layout to make sure they don't get loaded twice
- fix console errors for when there isn't a nav on page
- add toggle component (html,css,js)
- add new currentColor icon class
- add new icon class for flipped icon
- add toggle documentation

### DIFF
--- a/packages/spark-core/components/_icons.scss
+++ b/packages/spark-core/components/_icons.scss
@@ -2,17 +2,24 @@
 // Icons
 // ==================================================================
 .sprk-c-Icon {
-  height: $font-size-body-two;
+  height: $icon-m;
   fill: currentColor;
-  width: auto;
+  width: $icon-m;
 }
 
-.sprk-c-Icon--small {
-  height: 0.824rem;
+.sprk-c-Icon--l {
+  height: $icon-l;
+  width: $icon-l;
 }
 
-.sprk-c-Icon--large {
-  height: $font-size-body-two + 0.25;
+.sprk-c-Icon--xl {
+  height: $icon-xl;
+  width: $icon-xl;
+}
+
+.sprk-c-Icon--xxl {
+  height: $icon-xxl;
+  width: $icon-xxl;
 }
 
 .sprk-c-Icon--open {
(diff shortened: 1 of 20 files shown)


packages/spark-core/components/_icons.scss
```diff
@@ -13,3 +13,12 @@
 .sprk-c-Icon--large {
   height: $font-size-body-two + 0.25;
 }
+
+.sprk-c-Icon--currentColor {
```
yodasw16: Should this just be the default for any icon? Like `.sprk-c-Icon` has `fill: currentColor` set and then a particular icon might override it if needed, like `.sprk-c-Icon--exclamation { fill: red }`?

packages/spark-core/components/toggle.js
```diff
@@ -0,0 +1,63 @@
...
+ *  on load if we detect the toggle container.
+ */
+// Toggle the hide class on the content
+const toggleContentCSS = (toggleContent) => {
+  toggleContent.classList.toggle('sprk-u-Hide');
+};
+
+// Toggle the flip class on the icon
+const toggleIconCSS = (toggleIcon) => {
+  toggleIcon.classList.toggle('sprk-c-Icon--flip');
```
RCopeland: maybe not 'flip', since it might not always do that to indicate 'open'

packages/spark-core/components/toggle.js
```diff
@@ -0,0 +1,63 @@
...
+    const isExpanded = toggleTrigger.getAttribute('aria-expanded');
+    toggleTrigger.setAttribute('aria-expanded', isExpanded === 'false' ? 'true' : 'false');
+  } else {
+    // If it doesn't have it then set it to false initially
+    toggleTrigger.setAttribute('aria-expanded', 'false');
+  }
+};
+
+const toggle = () => {
+  let toggleContainer = document.querySelectorAll('[data-sprk-toggle="container"]');
```
RCopeland: can use getElements utility here

packages/spark-core/components/_icons.scss
afebbraro: I originally had it like that but added it as a modifier because I waa seeing examples on apps where they had the icon mostly different color than the text. I’d rather have it apart of the default icon class and sounds like you would too so I’ll revert it back to the way I had originally. Woot!

PR #137, opened 2018-03-28
remove icon file include in test pages since its in layout now

afebbraro:


afebbraro:
Reviewed in person with Rob

PR #138, opened 2018-03-28
refactored form-state-changer, phil review update for inputs, added testing, coverage

RCopeland:


packages/spark-core/base/_inputs.scss
```diff
@@ -96,7 +96,7 @@ textarea {
 
 .sprk-b-InputContainer > .sprk-b-Label,
 .sprk-b-InputContainer > label {
-  margin-bottom: $space-s;
+  margin-bottom: 12px;
```
yodasw16: Should we be doing this instead of using the spacing classes? (We are talking about this in person, but I already wrote it so I'm adding it)

PR #144, opened 2018-04-02
Utilities

yodasw16:
Opening this so I can get some feedback.


packages/spark-core/utilities/_utilities.scss
```diff
@@ -2,166 +2,301 @@
...
+
+// Alignment
+.sprk-u-AbsoluteCenter {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+// Typography
+.sprk-u-Hyphenate {
```
yodasw16: I'm going to move this to Scales.

packages/spark-core/utilities/_utilities.scss
```diff
@@ -2,166 +2,301 @@
...
+.sprk-u-ScreenReaderText {
+  position: absolute;
+  left: -10000px;
+  top: auto;
+  width: 1px;
+  height: 1px;
+  overflow: hidden;
+}
+
+.sprk-u-ScreenReaderText.sprk-u-ShowOnFocus:focus {
```
yodasw16: The `showOnFocus` class is completely useless without `ScreenReaderText`. Do you think it's necessary to qualify it like this? Basically prevents it from being used without the other, but is that necessary?

packages/spark-core/utilities/_utilities.scss
```diff
@@ -2,166 +2,301 @@
...
-  background-color: $white;
+  background-color: $white !important;
+}
+
+//
+// Multi-property utilities
+///
+
+// Accessibility
+.sprk-u-ScreenReaderText {
```
yodasw16: I changed this from `VisuallyHidden` because the purpose of this is to make it readable by a screen reader, so I thought this name is more obvious what it's for.

packages/spark-core/utilities/_utilities.scss
```diff
@@ -2,166 +2,301 @@
...
+// Text transform
+.sprk-u-TextTransform--uppercase {
+  text-transform: uppercase !important;
+}
+
+.sprk-u-TextTransform--none {
+  text-transform: none !important;
+}
+
+// Text color
```
yodasw16: Not sure if we should have these color utilities at all. Text or background color. Background was that the highlight board component needs the ability to have different colored backgrounds and possibly different text colors based on the background color or background image. So I just went all in and created a class for everything.

Opinions? Maybe leave it out until it actually comes up again and we can figure out if it's the best way then?

packages/spark-core/utilities/_utilities.scss
RCopeland: probably dont need to qualify, somebody might make an override for sprk-u-ScreenReader or something similar and would need ShowOnFocus to be separate.

packages/spark-core/utilities/_utilities.scss
RCopeland: I'd rather see people using override classes in their app than using sprk-u classes for this. The potential for abuse with colors is too high, IMHO.

packages/spark-core/utilities/_utilities.scss
```diff
@@ -2,166 +2,301 @@
...
+//
+// These utilities should map directly to a single CSS property and
+// value and should always use !important.
+//
+// Properties with vendor prefixes count as single properties.
+///
+
+// Display and Visibility
+.sprk-u-Display--none {
+  display: none !important;
```
afebbraro: When this PR comes to a close (or maybe it should be a part of this PR) - I recommend going through and finding all the `display: none` places that are currently in the code and removing them and adjusting to use these.

RCopeland:
How are we going to document utilities? We'll also need a test page for them.

packages/spark-core/tools/_animations.scss
```diff
@@ -2,6 +2,10 @@
 // Animations
 // ==================================================================
 
+//
+// Infinite Rotation
+///
+
```
afebbraro: nice catch

packages/spark-core/utilities/_utilities.scss
```diff
@@ -2,175 +2,211 @@
...
 // ==================================================================
 
-// TODO: Dave is doing utils
-.sprk-u-Hide {
-  display: none;
+//
+// Single property utilities
+//
+// These utilities should map directly to a single CSS property and
+// value and should always use !important.
```
afebbraro: i appreciate the comment heading here!

PR #145, opened 2018-04-03
tables

RCopeland:


afebbraro:
I checked this out in Netlify and here on Github and didn't find any issues. I was trying to find out if we needed that text-align center declaration for the main Table. But, I found that Scales' is left aligning cell text when the colspan="1". In the customizable spirit of Scales, it almost feels like Scales should offer a way to change that value. Maybe be exposing a $text-alignment var on the main Scales Table class. LIke `.#{$scales-namespace}#{$scales-base-class-namespace}Table {
  width: 100%;
  text-align: $text-alignment;
}`. Another way could be to offer up a .Table--centered Scales class. It seems that more applications besides us would want to be able to change that value since centered text in Tables is common. @yodasw16  Thoughts?

PR #146, opened 2018-04-03
spinners

RCopeland:


src/patterns/components/spinners/dark.hbs
```diff
@@ -0,0 +1,8 @@
+---
+name: Dark Spinner
+notes: Dark variant is used when its needed to put a spinner against a light background.
```
afebbraro: For the notes: sentence here, I thought something like this might improve the sentence.

> The Dark Spinner variant is used to improve visibility when the spinner is placed against a light background color.

src/patterns/components/spinners/large.hbs
```diff
@@ -0,0 +1,8 @@
+---
+name: Large Spinner
+notes: The large variant is used in special cases.
```
afebbraro: Do we have any example's of special cases that would could detail out?

afebbraro:
General Question regarding Spinners: Do we have any rules about if Spinners can be used next to text? Say, something like` <button><spinner> Loading</button>`. We should add that stuff to the docs in descriptions.

packages/spark-core/tests/spinners.tests.js
```diff
@@ -0,0 +1,84 @@
+/* global document beforeEach describe it */
```
afebbraro: I didn't see an example of the spinner click event on the test page or the drizzle docs page. Might be worthwhile to add it in so apps could see the supplied spinner JS in action.

src/patterns/components/spinners/large.hbs
RCopeland: @philrinke thoughts?

packages/spark-core/tests/spinners.tests.js
RCopeland: the only one is on the buttons test page, for submission spinner

packages/spark-core/components/_spinners.scss
```diff
@@ -19,10 +19,14 @@ $spinner-color: $white !default;
 .sprk-c-Spinner--circle {
   border: $spinner-thickness solid $spinner-color;
   border-radius: 50%;
-  border-right-color: transparent;
+  border-right-color: transparent !important;
```
yodasw16: why does this need important?

PR #157, opened 2018-04-09
Modals and Issue #166

afebbraro:
- [x] Handle sending focus to last active element before opening modal upon closing of modal that was previously open
- [x] Handle tabbing (focus trapping) 
- [x] Modal HTML in the DOM above the closing body tag
- [x] Create three modal hbs files
- [x] Create test page
- [x] Add Close 'X' Icon to info modal
- [x] Allow user to 'ESC' from modal
- [x] Add data attributes for Cancel, Modal, and Mask
- [x] If modal mask is clicked when a modal is open then close the modal and send focus back
- [x] Add `aria-labelledby="modal-<modalName>-heading"` and `aria-describedby="modal-<modalName>-content"` to modal HTML
- [x] Add way to have multiple modals on the page at one time and still have JS work by using `data-sprk-modal-trigger-<unique-name>` and having the JS grab the unique name from the trigger and search for the corresponding modal
- [x] Have data attribute for modal cancel button and it closes modal if clicked
- [x] When Wait modal is open, clicking on the modal mask or hitting ESC should not close the modal
- [x] Talk with @RCopeland  about the "Button next to link" layout pattern
- [x] Talk to Phil about adding a close icon to the "wait" modal
- [x] Hitting enter on the x icon closes the modal
- [x] Write modal unit tests
- [x] Separate modal JS out into function utilities and one that is modal specific
- [x] Expose functions for modals, basic thing that apps can use if they want, they would have to add the data-attributes.
- [x] Prevent scrolling when modal is open
- [x] If modal content is too tall on landscape phone screen - Allow Scroll Bar on modal content
- [x] Wait modal type has no focusable elements so make sure the modal container is focus trapped
- [x] Add focusable="false" to svg close icons that are inside the <button> because ie11 sends focus there when tabbing
- [x] Testing in IE, Edge, FF, Chrome, Safari ETC
- [x] Remove outline from modal because in this context(modal container itself not focusable) 
- [x] Modal Drizzle Documentation
- [x] Finalize modal HTML/SCSS


afebbraro:
@RCopeland  and @yodasw16, I realized one more thing I needed. Remove mask, and key event listeners when modal is closed. This can still be reviewed in the mean time though.

packages/spark-core/utilities/elementState.js
```diff
@@ -0,0 +1,20 @@
+/* global document */
+// Get all focusable elements in a container
+const getFocusableEls = (container) => {
+  let focusEls = container.querySelectorAll('a[href], area[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), button:not([disabled]), [tabindex="0"]');
+  focusEls = Array.prototype.slice.call(focusEls);
```
RCopeland: is the Array conversion necessary?

packages/spark-core/package.json
```diff
@@ -23,6 +23,7 @@
     "@scalescss/scalescss": "^3.4.0",
     "chai": "^4.1.2",
     "jsdom": "^11.6.2",
+    "npm": "^5.8.0",
```
RCopeland: i dont think we want to upgrade npm

PR #202, opened 2018-05-03
Modal Scroll: trying stuff in staging pr preview on mobile

afebbraro:


packages/spark-core/components/modals.js
```diff
@@ -114,8 +114,8 @@ const showModal = (modal, mask, main) => {
   main.setAttribute('aria-hidden', 'true');
 
   // Prevent background from scrolling
-  document.body.classList.add('sprk-u-OverflowHidden');
   document.documentElement.classList.add('sprk-u-OverflowHidden');
+  document.body.classList.add('sprk-u-OverflowHidden');
```
yodasw16: Did this really end up working?

afebbraro: @yodasw16 haha yeah but it makes the page scroll to the top and it stays there when the model closes. 😑

PR #211, opened 2018-05-08
narrow nav refactor

RCopeland:


src/patterns/components/navigation/narrow-navigation.hbs
```diff
@@ -5,65 +5,58 @@ description: |
...
-          <div class="sprk-o-SideBySide__item sprk-o-SideBySide__item--three-quarters">
-            <span>Item 1 </span>
-          </div>
-          <div class="sprk-o-SideBySide__item sprk-o-SideBySide__item--one-quarter">
-            <svg class="sprk-c-Icon sprk-c-NarrowNavigation__arrow" viewBox="0 0 100 100" data-sprk-toggle="icon">
-              <use xlink:href="#chevron-down" />
-            </svg>
-          </div>
-        </div>
+<nav>
```
afebbraro: Should you add role="navigation"?

src/patterns/components/navigation/narrow-navigation.hbs
```diff
@@ -5,65 +5,58 @@ description: |
...
+  <ul class="sprk-c-Accordion sprk-c-Accordion--navigation sprk-b-List--bare">
```
afebbraro: I think you forgot the sprk-b-List before the sprk-b-List--bare

src/patterns/components/navigation/narrow-navigation.hbs
```diff
@@ -5,65 +5,58 @@ description: |
...
-          </div>
-          <div class="sprk-o-SideBySide__item sprk-o-SideBySide__item--one-quarter">
-            <svg class="sprk-c-Icon sprk-c-NarrowNavigation__arrow" viewBox="0 0 100 100" data-sprk-toggle="icon">
-              <use xlink:href="#chevron-down" />
-            </svg>
-          </div>
-        </div>
+
+    <li class="sprk-c-Accordion__item {{item}}" data-sprk-toggle="container">
+      <a aria-controls="details1" class="sprk-c-Accordion__summary {{summary}}" data-sprk-toggle="trigger" data-sprk-toggle-type="accordion" href="#">
```
afebbraro: Something we should figure out in the future just to clean the code blocks up in the docs is the {{vars}} that we use. Right now if the component doesn't have that set then it leaves an extra white space in our doc output like `class="sprk-c-Accordion__summary  "` which, isn't the biggest deal, of course. But, it is something we can avoid. But, would take writing extra code. I think I did it for one component by using if statements to check if it's there. But, I didn't do it for every component I've wrote. Anyway! Thought I'd bring it up.

PR #257, opened 2018-05-24
Suggestion for content edit: clarify the placement for TypeKit

erikwoods:
Today, it says "before the closing tag." That's not very specific. So, making some assumptions, I've provided a suggestion for editing that content.


afebbraro:
thank you @erikwoods !

PR #260, opened 2018-05-24
input-fixes

RCopeland:


RCopeland:
**API changes**
- data-sprk-input="datepicker" becomes data-sprk-datepicker
- date input now invalidates unreasonable dates ( months higher than 12, days higher than 31, etc.)
- input fields that have formatting / validation (SSN, phone, monetary, date), now all validate and format on 'input' and 'blur' (except monetaryInput, which only validates on 'input', but still validates and formats on 'blur'.)

packages/spark-core/components/pagination.js
```diff
@@ -1,64 +1,28 @@
-/* global document */
+import getElements from '../utilities/getElements';
```
afebbraro: Nice! This reduces a lot of code.

packages/spark-core/tests/isValidSSN.tests.js
```diff
@@ -44,4 +44,12 @@ describe('isValidSSN tests', () => {
   it('should fail on scientific notation', () => {
     expect(isValidSSN('1.23e10')).eql(false);
   });
+
+  it('should fail on mismatch separators', () => {
+    expect(isValidSSN('333-33 3333')).eql(false);
+  });
+
+  it('should pass on with dashes', () => {
```
afebbraro: "should pass on with dashes" should be "should pass with dashes"

PR #268, opened 2018-05-31
Why just be good?

scottux:
Fixing the grammar a bit, then I thought our design adjective was setting a low goal so let's aim for greatness :)


afebbraro:
@scottux thank you!

PR #271, opened 2018-06-01
z-index-system

RCopeland:


src/patterns/fundamentals/layers/base.hbs
```diff
@@ -0,0 +1,83 @@
...
+          <li>Choice Modal (Overlay)</li>
+          <li>Wait Modal (Overlay)</li>
+          <li>Info Modal (Overlay)</li>
+        </ul>
+      </td>
+    </tr>
+    <tr>
+      <td>
+        <p class="sprk-u-FontWeight--bold">$layer-height-xl</p>
+        <p>The uppermost layer. No elements should occupy be positioned above this level.</p>
```
afebbraro: "No elements should occupy be positioned above this level" - could be reworded a bit. Maybe remove the word occupy?

scottux:
It should be very rare that you would need z-indexes this large. Can you explain this system a bit?

I would imagine there being a way to define a parent-child relationship such that the datepicker's container would be position relative and then the datepicker itself would only need to have a z-index of 1 to appear correctly.

Is there a "do not position things unless you want z-indexing to break" rule we should be following such that only the body gets the parent positioning?

scottux:
Just trying to consider the [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context)

RCopeland:
@scottux Hey, the idea is that any elements that need z-index should choose one of the provided variables. We chose the actual z-index values kind of arbitrarily, with the idea that it would be better if a legacy app's z-indexes are overpowered by Spark's. Now that I'm rereading it, maybe we don't need to specify the base level since there's no action to take.

PR #275, opened 2018-06-04
Add Bug Report Issue Template

afebbraro:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,35 @@
+---
+name: Bug report
+about: Create a report to help us improve
+
+---
+
+**Describe the bug**
```
yodasw16: Can you make these headlines instead? When I preview it, the bold lines and the line under it end up on the same line.

PR #288, opened 2018-06-06
Unit test coverage bump (up to 100%)

RCopeland:


packages/spark-core/tests/modals.tests.js
```diff
@@ -263,6 +478,120 @@ describe('Modal tests', () => {
...
+    showModal(defaultModal, modalMask, main);
+    link2.focus();
+    handleModalKeyEvents(modalsList, modalMask, main, tabKeyEvent);
+    expect(document.activeElement).eql(link2);
+  });
+
+  it('should do nothing if tab is pressed while the focused element is not last', () => {
+    const modalsList = document.querySelectorAll('[data-sprk-modal]');
+    const tabKeyEvent = {
+      key: 'Tab',
```
afebbraro: now that we are using sinon - maybe these event vars should be switched to the new way

PR #369, opened 2018-06-22
Example two

RCopeland:


src/pages/tests/example-two.hbs
```diff
@@ -0,0 +1,322 @@
+--- title: Page Example Two layout: blank ---
```
afebbraro: I see this being rendered as text in the top of the page. Was that what you wanted it to do?

afebbraro:
Whever I clicked loan summary in the narrow nav, it highlights the "todo" nav item and the loan summary. 
screen shot 2018-06-22 at 12 54 34 pm

PR #374, opened 2018-06-26
Angular setup

RCopeland:


src/angular/src/app/button-docs/button-docs.component.ts
```diff
@@ -0,0 +1,15 @@
+import { Component, OnInit } from '@angular/core';
+
+@Component({
+  selector: 'app-button-docs',
+  templateUrl: './button-docs.component.html',
```
afebbraro: Thoughts on using template instead of templateUrl? I think having the html in the component ts file would make development easier (to associate the ts with the html) and cut down on the number of files Spark has to have.

src/angular/src/app/button-docs/button-docs.component.ts
```diff
@@ -0,0 +1,15 @@
...
+  styleUrls: ['./button-docs.component.css']
```
afebbraro: We can get rid of all the .css files. If they want to add additional css overrides they can use 'style' here in the .ts file and write the code there. They wouldn't need an additional .css file.

PR #411, opened 2018-08-07
Bug fixing

RCopeland:


packages/spark-core/components/_accordions.scss
```diff
@@ -32,9 +32,7 @@
   text-decoration: none;
 
   &:hover,
-  &:active,
-  &:focus,
-  &:visited {
+  &:active {
```
yodasw16: should probably have some sort of focus state on this. Also, does removing visited instead of styling it make the default visited state bleed through? I don't know what that would actual look like.

RCopeland: it still has the blue outline for focus, just not the blue background / white color anymore. 
I'll test it across browsers for the visited color coming through.

RCopeland: Tested across browsers, no bleeding. I think its because the color of the text in the link is actually controlled by the h3 thats surrounding it. The bug was there because we were applying a color change to the h3's class when the outer link was :visited

afebbraro:
screen shot 2018-08-13 at 11 12 33 am
I'm noticing this on the deploy preview in chrome with this change. Click the nav accordion items and then unclick, click again, ETC. You'll notice the font color is white on the white bg.

PR #412, opened 2018-08-08
dictionary

RCopeland:


src/angular/src/app/spark-core-angular/components/sprk-dictionary/sprk-dictionary.component.ts
```diff
@@ -0,0 +1,39 @@
...
+  @Input() dictionaryType: string = 'base';
+  @Input() additionalClasses: string;
+  objectKeys: Function = Object.keys;
+
+  getClasses(): string {
+    let classArray: Array<String> = [
+      'sprk-c-Dictionary'
+    ];
+
+    if (this.dictionaryType === 'striped') {
```
afebbraro: This seems good because if they put the value of dictionaryType as "basic", which is the default, it will still work.

PR #429, opened 2018-08-20
Updates angular

RCopeland:


src/angular/src/app/app.component.scss
```diff
@@ -39,6 +39,7 @@
 @import '../../../../packages/spark-core/base/inputs';
 @import '../../../../packages/spark-core/base/lists';
 @import '../../../../packages/spark-core/base/links';
+@import '../../../../packages/spark-core/base/tables';
```
afebbraro: You just need to remove the unneeded table scss import here since it was covered in my branch that was merged.

Issue #443, opened 2018-08-27
[DOCUMENT] Inputs Documentation Updates


bobwise:
Fixing some typos and rewriting some of the item descriptions for clarity.

bobwise:
Moving this back to the backlog. I want to revisit how we're handling the <code> element throughout the documentation.

PR #447, opened 2018-08-28
Doc updates - Inputs

bobwise:
Issue https://github.com/sparkdesignsystem/spark-design-system/issues/443

Adding a class for inline code and revisiting the documentation for Inputs.

Information, Restrictions, and Links on individual components were rewritten, but they won't render in the app until PR https://github.com/sparkdesignsystem/spark-design-system/pull/445 is merged.


src/assets/drizzle/styles/components/_command.scss
```diff
@@ -10,3 +10,8 @@
 .#{$app-namespace}-c-Command code {
   font-size: 0.9rem;
 }
+
+.#{$app-namespace}-c-Command--inline {
+  padding: 0 0.25rem;
+  font-size: 0.9rem;
```
afebbraro: BEM modified classes utilize all the classes of the parent. So, this should be structured so that the element gets all of the original c-Command styles. That would make the font-size declaration in c-Command--inline not needed.

bobwise: c-Command doesn't actually have a font-size. There is a "c-Command code" selector that has the same font size, but that selector doesn't apply to the places where I'm using c-Command--inline. 

I didn't want to mess with the "c-Command code" selector because I didn't want to have to rewrite all of the places where it's used, but I can do that if it's the right thing to do.

afebbraro: I think we should revisit it because it’s weird the way we have it setup now. Shouldn’t be selecting elements directly. I’d like to know why we have the c-Command code instead of a modified class.

yodasw16: I wonder if the `<code>` element is being generated or something, making it hard to put a class directly on it. I remember being really confused on how to use that class properly, so I support revisiting.

bobwise: I'll revisit all the places where we're using <code>.  We also have a 'drizzle-CodeHighlight' class on a bunch of code elements that doesn't appear to be used in any CSS selectors.

PR #461, opened 2018-08-30
Add Fall Buttons, Links, and colors.

afebbraro:
This includes rebasing the fall branch with Master as well.


packages/spark-core/components/_buttons.scss
```diff
@@ -3,6 +3,7 @@
 // ==================================================================
 .sprk-c-Button {
   transition: background-color $btn-transition-speed;
+  border: 0;
```
afebbraro: Primary buttons do not have a border.

packages/spark-core/components/_buttons.scss
```diff
@@ -21,42 +22,46 @@
...
 }
 
-.sprk-c-Button--removal {
+.sprk-c-Button--tertiary {
   background-color: transparent;
-  border: 2px solid $red-medium;
-  color: $red-medium;
+  border: $btn-border-width solid $black;
+  color: $black;
+  padding: ($space-m - $btn-border-width) $space-l;
```
afebbraro: Because tertiary buttons have a border, we have to account for that space to ensure they are the same height as regular buttons.

afebbraro:
re-went through and updated the border-width from Brendan and went through code with RV. 😃 Thanks!

RVMendoza:
3o7abkhopu0nwenh3o
Checked over this with @afebbraro and this looks goooooood!

PR #463, opened 2018-08-30
Update alerts

afebbraro:


src/pages/gettingstarted/developers.hbs
```diff
@@ -20,6 +20,15 @@ title: Developers
       and polyfill appropriately for browser support.</p>
   </div>
 
+  <div class="drizzle-o-ContentGrouping">
+    <h2 class="drizzle-b-h2">Theming</h2>
+    <p>
+      Spark components use Sass variables inside a _settings.scss file for things like color,
```
yodasw16: I'd like to hold off on this. We need to be clear about when/if someone should be overriding the Spark variables. I would rather wait to add documentation about theming until we have design tokens, but if it goes in here now, it needs to be expanded.

afebbraro: I agree with that. I’ll remove it from this PR and add an issue for it to be done in a more complete way.

PR #465, opened 2018-08-30
Fall inputs

RCopeland:


packages/spark-core/_spark-core.scss
```diff
@@ -45,7 +46,12 @@
 @import './tools/animations';
 @import './tools/functions';
 @import './tools/mixins';
-@import './base/inputs';
+@import 'base/inputs/error-container';
```
afebbraro: Thinking maybe we should have a manifest for all the inputs so we don't have to import them all separately like this (since core stuff is made to go together anyway).

packages/spark-core/base/inputs/_error-container.scss
```diff
@@ -0,0 +1,17 @@
...
+  margin-top: $space-s;
+}
+
+.sprk-b-ErrorIcon {
+  fill: $text-input-error-color;
+  margin-right: $space-s;
+}
+
+.sprk-b-ErrorText {
+  color: $black;
```
afebbraro: Does Error text need this font color declaration since this will be the default type color?

packages/spark-core/base/inputs/_text-input.scss
```diff
@@ -0,0 +1,97 @@
...
+  position: relative;
+}
+
+.sprk-b-DatePicker__icon,
+.sprk-b-TextInput--monetary::before {
+  position: absolute;
+  z-index: $layer-height-xs;
+}
+
+.sprk-b-DatePicker__icon {
```
afebbraro: It's not related to this line but wanted to note that on the iPhone, when you click into the datepicker field (regardless of how far up or down the datepicker is on your screen) It opens the datepicker and also opens the keyboard which flys up and you cant see the datepicker to pick a date. If you close the keyboard to pick a date then the datepicker closes. So, you cant pick a date on iPhone.

afebbraro: i noticed the "copy" code button is covering the datepicker as well. I have an issue out that fixes this for modals by adding a higher z-index for modals in the spark site. Thinking maybe we need to do that for datepickers in the spark site or (better yet) change the elevation of the copy code button in the spark site.

packages/spark-core/base/inputs/_text-input.scss
```diff
@@ -0,0 +1,97 @@
...
+  margin-top: $input-icon-offset-y;
+  margin-left: $input-icon-offset-x;
+}
+
+.sprk-b-TextInput--monetary::before {
+  content: '$';
```
afebbraro: I had an idea for an enhancement for theming on this. Could make the '$' a scss var. So, users who are using a different currency than ours or for whatever reason they can put whatever they want there.

packages/spark-core/base/inputs/_text-input.scss
```diff
@@ -0,0 +1,97 @@
...
+  display: block;
+  font-size: 1rem;
+  margin-top: $input-text-icon-offset-y;
+  margin-left: $input-text-icon-offset-x;
+  line-height: 1;
+}
+
+.sprk-b-TextInput[readonly] {
+  color: $white;
```
afebbraro: This should be a var too.

packages/spark-core/base/inputs/_text-input.scss
```diff
@@ -0,0 +1,97 @@
...
+  font-weight: bold;
+}
+
+.sprk-b-TextArea {
+  min-height: 125px;
+  margin: $space-stack-m;
+}
+
+.sprk-b-Fieldset {
+  padding: 0;
```
afebbraro: Don't need this because Scales exposes the $fieldset-padding variable that we can set to 0 in our settings file.

packages/spark-core/base/inputs/_text-input.scss
afebbraro: Scales exposes $input-readonly-text-color for us to use.

packages/spark-core/base/inputs/_text-input.scss
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// Placeholder colors
+::-webkit-input-placeholder,
+::-moz-placeholder,
+:-ms-input-placeholder,
+::placeholder {
+  color: $gray-light;
```
afebbraro: We should expose placeholder color as a var.

packages/spark-core/settings/_settings.scss
```diff
@@ -251,18 +251,37 @@ $list-margin-left: $space-m !default;
 //
 // Forms
 ///
-$text-input-border-radius: $border-radius !default;
-$text-input-border-color: $gray-medium !default;
-$text-input-border-width: 2px !default;
+$input-gray-medium: #5a5a5a;
```
afebbraro: Forgot to add !default here.

packages/spark-core/settings/_settings.scss
```diff
@@ -251,18 +251,37 @@ $list-margin-left: $space-m !default;
...
+$input-icon-offset-y: 7px !default;
+$input-text-icon-offset-x: 4px !default;
+$input-text-icon-offset-y: 8px !default;
+$input-animation-speed: 0.3s !default;
+$select-arrow-offset-y: -23px !default;
+$select-arrow-offset-x: 8px !default;
+$label-offset-y: -50px !default;
+$label-offset-y-open: -31px !default;
+$label-counter-weight: 28px !default;
+$label-shrink-scale: 0.82;
```
afebbraro: Forgot to add !default here.

src/assets/toolkit/styles/toolkit.scss
```diff
@@ -36,7 +37,12 @@
 @import '../../../../node_modules/@scalescss/base-tables/tables';
 @import '../../../../packages/spark-core/base/tables';
 @import '../../../../packages/spark-core/base/typography';
-@import '../../../../packages/spark-core/base/inputs';
+@import '../../../../packages/spark-core/base/inputs/error-container';
```
afebbraro: I think this would be easier if we had the one input manifest.

packages/spark-core/base/inputs/_input-container.scss
```diff
@@ -0,0 +1,21 @@
+.sprk-b-InputContainer {
+  max-width: $input-max-width;
+  margin-top: 2.5rem;
+}
+
+.sprk-b-InputContainer\@sm {
```
afebbraro: The format for responsive suffixes is `.thing--what-it-does@breakpoint`. For example, our buttons are `sprk-c-Button--full\@sm`. It's saying the buttons are full at the small breakpoint. I searched this PR for this selector and only saw it here. Maybe its some dev code that you forgot to remove?

packages/spark-core/base/inputs/_text-input.scss
```diff
@@ -0,0 +1,97 @@
+// ==================================================================
+// Input Component
+// ==================================================================
+.sprk-b-TextInput {
+  font-size: $font-size-body-two;
+  font-weight: $font-weight-body-two;
+  line-height: $line-height-body-two;
+  outline: 0;
+  padding: 3px $space-s;
```
afebbraro: Scales has $text-input-padding that sets this for this class.

packages/spark-core/base/inputs/_text-input.scss
```diff
@@ -0,0 +1,97 @@
...
+  box-shadow: $input-box-shadow;
+}
+
+.sprk-b-TextInput[disabled] {
+  background-color: $gray-light;
```
afebbraro: Scales has  $input-disabled-background-color and $input-disabled-text-color for us to use.

packages/spark-core/base/inputs/_error-container.scss
RCopeland: ill move it out to a var

packages/spark-core/base/inputs/_input-container.scss
```diff
@@ -0,0 +1,21 @@
+.sprk-b-InputContainer {
+  max-width: $input-max-width;
+  margin-top: 2.5rem;
```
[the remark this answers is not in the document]
RCopeland: There's some margin coming from scales that i didnt see. I'll override it to only have the margin thats necessary to make room for labels.

packages/spark-core/base/inputs/_input-container.scss
RCopeland: Its code that was there before (I think), which isnt necessary now, I'll remove it.

packages/spark-core/base/inputs/_text-input.scss
RCopeland: hmm, might be best to disable the picker on mobile, the field functions as a normal date field anyway.

RCopeland: The zindex system we're using now puts all drizzle stuff higher than spark component stuff. That might not be the best approach. I'll take a look, will likely fix in a separate PR though.

RCopeland: will address datepicker changes in another pr

PR #472, opened 2018-08-31
Add pagination fall

afebbraro:


packages/spark-core/components/_pagination.scss
```diff
@@ -8,33 +8,71 @@
...
+    border-color: $pag-selection-color;
   }
 }
 
-// For items in the long pagination version
-.sprk-c-Pagination__item--long {
-  margin-right: $space-xs;
+.sprk-c-Pagination__link--current {
+  background-color: $pag-selection-color;
+  color: $white;
```
RCopeland: this should be a var too probably, cause if they can set $pag-selection-color to something else, they might need to change the color too.

afebbraro: ah yeh! Good thinking.

PR #486, opened 2018-09-06
Fall dictionary

RCopeland:


packages/spark-extras/components/dictionary/_dictionary.scss
```diff
@@ -5,17 +5,18 @@
...
-.sprk-c-Dictionary--striped {
+.sprk-c-Dictionary {
   border: $dictionary-border;
+  padding: $space-m;
 }
 
 .sprk-c-Dictionary__keyvaluepair {
   width: 100%;
-  padding: $space-inset-m;
+  padding: $space-m / 2;
```
afebbraro: Why the division by 2? In this case, 16 / 2 is 8, which is $space-s.

packages/spark-extras/components/dictionary/_dictionary.scss
```diff
@@ -24,22 +25,34 @@ $dictionary-breakpoint: 38.4375rem !default;
   }
 }
 
+.sprk-c-Dictionary .sprk-b-Label {
+  font-size: $font-size-body-four;
```
afebbraro: The key is smaller than the value. Is that supposed to be like that?

packages/spark-extras/components/dictionary/_dictionary.scss
RCopeland: in the case of the striped dictionary, i need to have space-m space between, but divided between the top and bottom. I could use space-s instead, but if space-s changes and space-m doesnt it would break. (Not sure thats a real thing that would happen)

packages/spark-extras/components/dictionary/_dictionary.scss
RCopeland: by design

afebbraro: lol the only time i've use the thumbs down 😩

PR #489, opened 2018-09-07
Fall Accordions

janinavelasco9:
This is for primary navigations right now.
Will add secondary navigation when we get design direction.


src/assets/toolkit/images/icons/chevron-down-circle-filled.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 25 25"><circle class="sprk-SvgChevronCircleFilled--circle" cx="12.46" cy="12.54" r="11.2"/><polyline class="sprk-SvgChevronCircleFilled--arrow" points="18.36 10.64 12.46 16.54 6.56 10.64"/></svg>
```
yodasw16: @afebbraro @RCopeland is this why we are getting mixed content warnings? I can't see anything else that would cause it. I wouldn't expect this to cause it either, but I dunno.

RCopeland: i _think_ its from a placeholder image on the links page
https://github.com/sparkdesignsystem/spark-design-system/pull/527

yodasw16: I commented here because the email directed me to a netlify log that pointed back to this PR, not that links page. https://app.netlify.com/sites/spark-design-system/deploys/5b9ab1db1f12b742318dac61

packages/spark-core/components/toggle.js
```diff
@@ -46,6 +47,23 @@ const handleToggleClick = (toggleContent, toggleIcon, element) => {
   toggleAriaExpanded(element);
 };
 
+const handleToggleChevronIcon = (document, element) => {
+  // If the element contains 'sprk-c-Accordion__summary--open',
+  // set xlink:href attribute to be the filled icon and add a class to change
+  // styles.
+  const sprkChevronCircle = element.querySelector('use');
+  if (element.classList.contains('sprk-c-Accordion__summary--open')) {
+    const circle = document.querySelector('.sprk-SvgChevronCircleFilled--circle');
```
afebbraro: Hey @janinavelasco9, wanted to give a bit of feedback before you got too far into unit tests. You'll want to do the query selection inside the bindToggleUIEvents. That away, you already have access to the accordion element and aren't running a query on the entire document again. So, you can more the const circle... and const arrow... into there. That will make the unit tests much more easy as well. Hope thats helpful.

afebbraro: You might be able to utilize the toggleIconCSS function for adding classes as well. Nice job so far and thanks again for helping out!

janinavelasco9: Yeahhh I didn’t push it up yet but I called that function in the click event in the bindtoggleuievent. But that’s a good idea to check out the toggle iconcss function. I’ll play around with that Monday. But for sureeee anything else you see needs work, let me know :) learning a lot.

afebbraro: Yay! That’s great to hear. 😃

packages/spark-core/components/_accordions.scss
```diff
@@ -12,46 +12,83 @@
...
   display: flex;
-  padding: $space-m;
+  padding: $space-m $space-s;
   text-decoration: none;
 
   &:hover,
   &:active {
-    background-color: $blue-dark;
-    color: $white;
+    color: $red-real;
```
afebbraro: We'll want these colors set to vars and put in settings.scss so people can theme the component if needed.

packages/spark-core/components/_accordions.scss
```diff
@@ -12,46 +12,70 @@
...
+  &:active {
+    border-bottom: 2px solid $gray-medium;
+    background-color: $blue-dark;
+    color: $white;
+  }
 }
 
 // When open, add same style as when focused
 .sprk-c-Accordion__summary--open {
+  font-weight: bold;
```
afebbraro: `sprk-c-Accordion__summary--open` doesnt need the `font-weight: bold` declaration because the heading is already set to bold. If we want to ensure that the text is bold in the case someone decides not to use our ` <h3>` (shouldn't be the case but just trying to help explain my thoughts here) then the ` font-weight: bold` declaration should be on the `sprk-c-Accordion__heading` class that we created for the heading in the summary.

PR #491, opened 2018-09-07
Couple card touchups while we wait

afebbraro:
Add box shadow and light border to cards. Pretty sure they will need them.


packages/spark-core/settings/_settings.scss
```diff
@@ -207,7 +207,7 @@ $alert-color: $white !default;
 $alert-color-info: $blue !default;
 $alert-color-success: $green !default;
 $alert-color-fail: $yellow !default;
-$alert-shadow: 0 4px 5px $gray-light !default;
+$alert-shadow: 0 2px 2px rgba(0, 0, 0, 0.2) !default;
```
afebbraro: one alert touchup too

PR #493, opened 2018-09-10
Angular docs accordions updates

RCopeland:


packages/spark-core/components/_accordions.scss
```diff
@@ -26,6 +26,7 @@
 }
 
 .sprk-c-Accordion__summary {
+  align-items: center;
```
afebbraro: Is this old code updates? Or fall?

RCopeland: old, its a fix for aligning the old style accordions

PR #519, opened 2018-09-12
divider docs

RCopeland:


src/patterns/components/dividers/angular/info/base.hbs
```diff
@@ -0,0 +1,21 @@
...
+    <th>Input</th>
+    <th>Type</th>
+    <th>Description</th>
+  </tr>
+  </thead>
+  <tbody>
+  <tr>
+    <td class="sprk-u-FontWeight--bold">additionalClasses</td>
+    <td>string</td>
+    <td>Allows a string of classes to add, in addition to the spark classes. Intended for overrides.</td>
```
afebbraro: We should go through and make sure we are capitalizing Spark when we are done with docs.

PR #524, opened 2018-09-13
add angular pagination docs

afebbraro:


src/patterns/components/pagination/angular/code/long.hbs
```diff
@@ -0,0 +1,20 @@
...
+  prevLinkText="Prev"
+  [currentPage]="longConfig[0].currentPage"
+  [totalItems]="longConfig[0].totalItems"
+  [itemsPerPage]="longConfig[0].itemsPerPage"
+  (previousClick)="yourGoBackFunction($event)"
+  (pageClick)="yourGoToPageFunction($event)"
+  (nextClick)="yourGoForwardFunction($event)">
+</sprk-pagination>
+
+longConfig = [
```
RCopeland: this might be confusing, people might think there's like a config input

afebbraro: Ahh yeah, I can see that, I'll remove it. Good catch!

PR #528, opened 2018-09-14
add angular docs highlight board

afebbraro:


src/patterns/components/highlight-board/angular/code/full-width.hbs
```diff
@@ -0,0 +1,9 @@
+<sprk-highlight-board
+  heading="Lift the burden of getting a home loan"
+  subHeading="Fast, Powerful and Completely Online"
+  ctaText="Learn More"
+  ctaTwoText="CTA Two"
```
RCopeland: is this supposed to be ctaText2 now?

PR #535, opened 2018-09-14
Master

RCopeland:


src/data/colors.yaml
```diff
@@ -29,10 +29,10 @@
     color: "#44235D"
     rgb: "rgb(68,35,93)"
     property: "$purple-dark"
-  - label: Purple
+  - label: Purple Medium
     color: "#551387"
     rgb: "rgb(85,19,135)"
-    property: "$purple"
+    property: "$purple-medium"
```
afebbraro: Good change, haha I was thinking, what do we even use purple for anymore?

PR #552, opened 2018-09-20
add angular docs password, radio, phone

afebbraro:


src/patterns/components/inputs/angular/info/phone.hbs
```diff
@@ -0,0 +1,36 @@
...
+  <colgroup>
+    <col class="sprk-u-Width-40">
+    <col>
+  </colgroup>
+
+  <thead>
+    <tr>
+      <th>Directive</th>
+      <th>Description</th>
+      <th>Type</th>
```
RCopeland: probably just need directive and description

src/patterns/components/inputs/angular/docs/phone.hbs
```diff
@@ -0,0 +1,4 @@
+<p class="sprk-u-mbm">
+  When the field includes a valid input, the sprkFormatterPhoneNumber formatter
```
RCopeland: i dont think its necessary to say, since its already covered in the codeinfo table

PR #553, opened 2018-09-20
Remove styling on html label tag

RVMendoza:
When consuming spark, there were some unexpected changes because spark is adding styles directly to HTML tags. This time it's on `<label>`.

This is a use case when this doesn't work because the app consuming it has custom labels that are inadvertently getting styles from spark. These labels are custom to the app and don't want any styling from spark on it.

Adding styling straight on to the HTML tag (in this case) takes away the option away for a consumer to make custom components without having to do an override.

I know this conversation is on going whether or not we be styling directly on tags but this is an example of what could go wrong as more apps consume the design system.


RVMendoza:
3ohhwo4pzdfaz2sadu

RVMendoza:
closing this and putting it on https://github.com/sparkdesignsystem/spark-design-system/pull/560

PR #630, opened 2018-10-03
Add info about adding spark-core to imports array

afebbraro:


afebbraro:
Need 2 reviews for master merge. @yodasw16 can you take a look when you get a chance?

PR #644, opened 2018-10-05
Config date picker

RCopeland:


packages/spark-core/package.json
```diff
@@ -26,6 +26,7 @@
     "chai": "^4.1.2",
     "jsdom": "^11.12.0",
     "nyc": "^11.9.0",
+    "proxyquire": "^2.1.0",
```
afebbraro: Why the new need for another dependency? Should this been in devDependencies?

packages/spark-core/spark-core.js
```diff
@@ -29,24 +29,27 @@ import './utilities/polyfills/classListSVG';
...
-tabs();
-toggle();
-spinners();
-WideNavigation();
-masthead();
-alerts();
-textInput();
-selectInput();
+const SparkCore = ({ datePickerConfig = {} } = {}) => {
+  detectJS();
```
afebbraro: If we are removing the Init fns then do we need to update the getting started developers page?

packages/spark-core/package.json
RCopeland: proxyquire lets unit tests attach stubs to internal dependancies. I'm using it to have more isolated unit tests. I originally had it in devDependencies, but the linter likes it better in dependencies, not sure why.

packages/spark-core/spark-core.js
RCopeland: yep, will do

PR #648, opened 2018-10-05
add tints/colors

afebbraro:


RVMendoza:
This looks pretty good! So this would be a breaking change right?

packages/spark-core/settings/_colors.scss
```diff
@@ -2,46 +2,40 @@
...
-$orange-medium: rgb(255, 155, 0);
-$orange-dark: rgb(230, 120, 0);
-
-// Neutrals
-$gray-light: rgb(229, 229, 229);
-$gray-medium: rgb(170, 170, 170);
-$gray-dark: rgb(90, 90, 90);
-$black: rgb(34, 34, 34);
+// Primary
+$red-real: rgb(200, 16, 46);
```
yodasw16: I don't understand the names in here. `red-real` stands out. All of the others are just one word colors with tint percentage numbers. That makes sense to me. The other one that doesn't fit is `gray-light`

afebbraro: @yodasw16  I responded to you via Chat.

PR #691, opened 2018-10-18
add grid

afebbraro:
- [x] double check tests
- [x] Angular components
- [x] Angular docs


RVMendoza:
Connected to story https://github.com/sparkdesignsystem/spark-design-system/issues/689

packages/spark-core/objects/_stack.scss
```diff
@@ -141,6 +192,46 @@
       margin-right: $stack-spacing-huge;
     }
 
+    > .sprk-o-Stack__item--sixth\@s {
+      flex: 0 0 16.666666667%;
```
yodasw16: Not suggesting we change anything, but have we thought about using calc for stuff like this?

src/patterns/fundamentals/grid/stack-object.hbs
```diff
@@ -0,0 +1,17 @@
+---
+name: Stack Object
+description: The Stack object lays items out in a column.
```
yodasw16: I think we need to expand this a little. Something about how the Stack Object is not a grid in itself, but a foundation that can be used to create a grid. I'm not exactly sure on wording, but calling this page "Grid" and then not really getting into what people expect from a grid until half way down the page might be confusing.

packages/spark-core/objects/_stack.scss
afebbraro: @yodasw16, interesting idea! No, I didn't think about using it. I strayed away from the abstraction because I thought seeing the output of --sixth vs. seeing `calc(...)` in the dev tools was more useful when putting things together.

src/patterns/fundamentals/grid/stack-object.hbs
afebbraro: Great point, I'll elaborate.

packages/spark-core/objects/_stack.scss
yodasw16: I'm not sure it even matter since the browser just rounds things how it wants to anyway.

Issue #693, opened 2018-10-19
Refactor input HTML and CSS into BEM implementation


RVMendoza:
## Proposal Title
Create a component for a text input with icon
- Inline Inputs
- Date inputs

Currently, the structuring and css is not correct. There are containers with only modifier classes. There are modifier styles that modify components incorrectly

## Type of change
Please select only 1.
* [ ] New Component
* [x] Enhancement to existing component
* [ ] Substantial documentation change
* [ ] Changes to the design or features of [sparkdesignsystem.com](http://sparkdesignsystem.com)

## Change category
Select as many as apply.
* [ ] **Visual** (Changes to the visual design of a component)
* [ ] **UX** (Change the user experience design of a component)
* [x] **Component** (Modifying the functionality of a component)
* [ ] **Editorial** (Changes to documentation or content style)
* [ ] **Code Style** (Changes to the way code is written or the component architecture)
* [ ] **Accessibility** (Substantial changes to a component's accessibility)

![image](https://user-images.githubusercontent.com/4342363/47238507-72e33d80-d3b0-11e8-9f8c-c169190e37f6.png)
![image](https://user-images.githubusercontent.com/4342363/47238528-855d7700-d3b0-11e8-8a9b-fd4bf0200603.png)

PR #695, opened 2018-10-22
Refactor input HTML and CSS into BEM implementation

RVMendoza:
Fixes https://github.com/sparkdesignsystem/spark-design-system/issues/693

Create a component for a text input with icon

Inline Inputs
Date inputs

Currently, the structuring and css is not correct. There are containers with only modifier classes. There are modifier classes that modify components incorrectly

### DIFF
--- a/packages/spark-core/base/inputs/_labels.scss
+++ b/packages/spark-core/base/inputs/_labels.scss
@@ -67,19 +67,17 @@
   transform: translateY($label-offset-y) scale($label-shrink-scale); // ratio of body-four to body-three
 }
 
-.sprk-b-TextInput--date-picker .sprk-b-TextInput--has-value ~ .sprk-b-Label,
-.sprk-b-TextInput--date-picker .sprk-b-TextInput[placeholder] ~ .sprk-b-Label,
-.sprk-b-TextInput--date-picker .sprk-b-TextInput--focusin ~ .sprk-b-Label,
-.sprk-b-TextInput--monetary .sprk-b-TextInput--has-value ~ .sprk-b-Label,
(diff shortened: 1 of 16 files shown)


packages/spark-core/base/inputs/_labels.scss
```diff
@@ -65,19 +67,17 @@
...
+// We need to abstract this so a label is not dependent on being adjacent to a text input
+.sprk-b-TextInputIconContainer .sprk-b-TextInput--has-value ~ .sprk-b-Label,
+.sprk-b-TextInputIconContainer .sprk-b-TextInput[placeholder] ~ .sprk-b-Label,
+.sprk-b-TextInputIconContainer .sprk-b-TextInput--focusin ~ .sprk-b-Label {
   padding-left: $space-s;
 }
 
-.sprk-b-TextInput--date-picker .sprk-b-Label {
+.sprk-b-Label--with-icon {
   padding-left: $space-l;
```
afebbraro: We typically follow the format “has-thing” instead of “with-thing”. I’m not sure if we have that documented anywhere though so this isn’t a mistake you made! We should get an issue open if we don’t. Please update this to match the others.

src/patterns/components/inputs/angular/code/datepicker.hbs
```diff
@@ -1,5 +1,9 @@
-<sprk-input-container iconContainerClasses="sprk-b-TextInput--date-picker">
-  <label sprkLabel>Date</label>
+<sprk-input-container iconContainerClasses="sprk-b-TextInputIconContainer">
+  <label
+    [additionalClasses]="sprk-b-Label--with-icon"
```
afebbraro: @robcopland I thought we got rid of additionalClasses on directives?

PR #718, opened 2018-10-25
add ng code for grid docs and fix grid docs typos

afebbraro:


src/patterns/fundamentals/grid/stack-object-options.hbs
```diff
@@ -223,14 +223,6 @@ order: 2
       </td>
     </tr>
 
-    <tr>
```
afebbraro: this was a duplicate line

PR #728, opened 2018-10-29
Fall tabbed nav

RCopeland:


packages/spark-core/components/_tabs.scss
```diff
@@ -15,55 +16,29 @@
 
 .sprk-c-Tabs__button {
   -webkit-appearance: none;
-  background: $white;
+  background: none;
```
yodasw16: This should be `background-color`, but I'm wondering if it should be set to $gray instead and remove the `background-color` from line 5. So, background colors are always set on the buttons themselves, instead of the background of the whole component bleeding through.

PR #736, opened 2018-10-30
Inline search reskin

RCopeland:


src/angular/src/app/spark-docs/input-docs/input-docs.component.ts
```diff
@@ -137,16 +137,22 @@ import { Component } from '@angular/core';
...
+              sprkLabel>
+              Search
+            </label>
+            <sprk-icon iconType="search" additionalClasses="sprk-b-InlineSearch__icon" sprk-input-icon></sprk-icon>
             <input
               name="inline_search_input"
-              type="search"
-              role="search"
+              class="sprk-b-TextInput--with-icon"
+              type="text"
```
yodasw16: Previous version has `type="search"` and `role="search"`. Is there a reason for that change? Same in the docs.

PR #748, opened 2018-10-31
<hr /> documentation update- better explanation & usage

katehyland:
Added a better (and updated) description of using < hr /> & example usage.


RCopeland:
lgtm, @RVMendoza, is the explanation here what you had in mind? For #748

RCopeland:
I'll look into the failed build, something on our end maybe

RCopeland:
@katehyland Is the site building locally for you? Netlify is throwing an error on drizzle's copy step.

RVMendoza:
nice work on this @katehyland! 

An explanation we should have as well:  if a user needs a line that's not necessarily a semantic `<hr />` but more of as a divider/separator, then use: 

```
<span class="sprk-c-Divider">
</span>
```

But if it's an actual semantic use of `<hr />` then use an `<hr />` tag.

Because not all horizontal lines are semantically `hr`s –– sometimes they're just dividers

I hope that makes sense. Let me know if I need to explain more!

src/patterns/components/dividers/collection.yaml
```diff
@@ -1,5 +1,25 @@
 description:
-  A Divider logically divides
-  sections of content. Hr HTML elements will
-  have style by default so no additional CSS class is needed.
+Using <hr/> to break up sections:  
+<hr /> stands for Horizontal Rule. 
+This is used primarily when you want to separate a page into sections. 
+It represents a thematic break between paragraph-level elements.
+
+In HTML4, the <hr /> tag was used to display as a horizontal rule or line.
```
RVMendoza: I think we can remove lines 7-10, just to speed up the consumption of information.

As much as ***I*** am interested in the history of `<hr />`, our documentation is about what something *does* rather than what it *used* to do.

We would do well to remove this paragraph and keep it to the most relevant information.

RCopeland:
@katehyland ok, i'll try to pull it down and see if i can duplicate.

RCopeland:
@katehyland sorry for the delay. I just took a look, the build error is because in the collection.yml file, you have html and non yaml spacing. You could either put just text in the yaml file, or maybe create another hbs file with order 0 that could have a doclayout instead. For that idea, take a look at: https://github.com/sparkdesignsystem/spark-design-system/blob/staging/src/patterns/components/icons/sizes.hbs as an example.

src/patterns/components/dividers/collection.yaml
```diff
@@ -1,7 +1,14 @@
-description:
-  A Divider logically divides
-  sections of content. Hr HTML elements will
-  have style by default so no additional CSS class is needed.
+name: Using Horizontal Rule (<hr />)
```
RCopeland:
```suggestion
```

RCopeland: We want to leave the name as the default 'Dividers'

PR #753, opened 2018-10-31
updating nvmrc for netlify

RCopeland:


RCopeland:
i think we should change it, netlify is trying to use 10.13 or something when it reads lts/*

RCopeland:
we probably dont want our build environment changing right when LTS is updated anyway

PR #764, opened 2018-11-01
Critical bugs masthead nav

RCopeland:


packages/spark-core/spark-core.js
```diff
@@ -25,6 +25,7 @@ import './utilities/polyfills/StringIncludes';
 import './utilities/polyfills/ArrayFind';
 import './utilities/polyfills/NodeListForEach';
 import './utilities/polyfills/classListSVG';
+import './utilities/polyfills/focusWithin';
```
RCopeland: @yodasw16 @afebbraro Do you guys think its worth it to load a polyfill for this?
If I dont, then in IE / Edge, when a user tabs into a wide-nav item's submenu, the wide-nav item wont be marked active. Mouse use is fine, this would only affect keyboard only.

yodasw16: They will still be able to see which submenu item they are currently on? If so, the submenu being visually under the main item should be enough to indicate which one is active. Unless I'm misunderstanding.

On the other hand, the polyfill is _tiny_.

packages/spark-core/utilities/polyfills/focusWithin.js
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable */
+(function(window, document){
```
afebbraro: Should we convert this to ES6 to match the rest of our JS?

RCopeland: ended up removing the polyfill

PR #765, opened 2018-11-01
Link formatting

RCopeland:


RCopeland:
I'll fix the disabled link visited issue, but lets have a design review around the standalone link issue.

PR #767, opened 2018-11-02
758 - IE11 - add stretch column, remove width on dictionary, set heig…

janinavelasco9:
…ht on accordion


src/patterns/components/awards/primary.hbs
```diff
@@ -26,7 +26,7 @@ hasAngularCodeInfo: true
     </a>
   </div>
 
-  <div class="sprk-o-Stack__item sprk-o-Stack__item--start-column" data-sprk-toggle="container">
+  <div class="sprk-o-Stack__item sprk-o-Stack__item--stretch-column" data-sprk-toggle="container">
```
afebbraro: Hey @janinavelasco9, thanks for the PR! I had one change request for this. Align stretch is the flexbox default and in order to avoid repeating, we can remove the Stack parent's center class (`sprk-o-Stack--center-column`) and then allow the items inside to be set individually. For each item we would add `sprk-o-Stack__item--center-column` but omit that for the final disclaimer element which we want to have the defaults. Then we wouldn't need the stretch classes.

janinavelasco9: yes! i actually saw that stretch was the default and was making a similar change but then had to run to the gym. but what you said makes perfect since! ill make the change.

packages/spark-core/components/_accordions.scss
```diff
@@ -81,6 +81,7 @@
 
 .sprk-c-Accordion--navigation .sprk-c-Accordion__summary {
   border-top: 0;
+  height: $icon-xl;
   min-height: $icon-xl;
```
afebbraro: Hey @janinavelasco9, I'm curious about what the reasoning was for having both a min-height and height on this?

src/patterns/components/awards/primary.hbs
afebbraro: Thanks! Appreciate the extra help so much!

packages/spark-core/components/_accordions.scss
janinavelasco9: IE11 doesn't respect min-height. so i had to add a height for IE11 to center the content.

janinavelasco9: basically min-height on a flex container won't apply to its flex items. so i put the height on there for IE11 to recognize the height. maybe there is a better way to go about it?

afebbraro: @janinavelasco9  i think we need that to be removed. @RCopeland  is going to take get with you. I gotta head out for the day.

afebbraro: Approving for @RCopeland to take over

RCopeland: @janinavelasco9 adding the height wont work for the use case where we add like a sentence to the summary. The text will just overflow its container : (

Issue #819, opened 2018-11-13
[BUILD] Link updates


afebbraro:
- [ ] Drop the pixel size of the link underline for all link styles to 1px
- [ ] Remove bold weight from inline link only
- [ ] Move link underline up some (sync with @bbrown7 on how far to move it up)

Issue #792, opened 2018-11-08
[BUG] When Link text wraps the underline only shows on last line of text


afebbraro:
## Describe The Bug
When Link text wraps the underline only shows on last line of text.

## To Reproduce
1. Go to links
2. Enter long link text 
3. See error

## Expected Behavior
Link underline should perform the same way that text-decoration performs where it underlines each line of text.

## Screenshots
![screen shot 2018-11-08 at 8 28 03 am](https://user-images.githubusercontent.com/2117593/48201455-3ad87600-e330-11e8-8177-1cb12c8ee37e.png)
## Desktop (please complete the following information):
 - OS: all
 - Browser all
 - Version all

## Smartphone (please complete the following information):
 - Device: all
 - OS: all
 - Browser all
 - Version all

PR #822, opened 2018-11-14
Updating link underlines

bobwise:
Issue https://github.com/sparkdesignsystem/spark-design-system/issues/792
and
Issue https://github.com/sparkdesignsystem/spark-design-system/issues/819

We want an underline solution that supports:

1. Allow for the underline and the link text to be different colors
2. Work on non-white backgrounds
3. Wrap with text
4. Scale with font size (thicker line for larger fonts)
5. Adjustable underline thickness
6. Adjustable underline location (the distance between the underline and the link text)

We are using the "border method", which meets all requirements except for # 6.

Box shadows can not be styled for :visited links due to security restrictions on the web.

image


packages/spark-core/base/_links.scss
```diff
@@ -4,63 +4,50 @@
...
   color: $link-color;
-  font-weight: $link-font-weight;
+  font-weight: $link-inline-font-weight;
   position: relative;
   text-decoration: none;
   transition: $link-transition;
+  box-shadow: inset 0 $link-underline-width $link-underline-color;
 
   &:visited {
     color: $link-visited-color;
```
bobwise: inline and standalone links are using this style for visited links (black text)

afebbraro: I tried some things this morning in an attempt to help since on Friday you mentioned that we can't change box-shadow on visited state. I tried using CSS vars here https://codepen.io/afebbraro/pen/GwMLxb?editors=1111 but had no luck. The var does change the box-shadow color on :hover but again, not on :visited. :(

PR #845, opened 2018-11-19
Fix duplicate/omitted data-id attributes

bobwise:
Issue https://github.com/sparkdesignsystem/spark-design-system/issues/846


yodasw16:
I think this should just be opened against master since it is just a documentation fix. Otherwise it has to wait for a release to update.

bobwise:
reopening against master

PR #906, opened 2018-12-06
moving polyfills out to another file, removing extraneous entries,

RCopeland:
updated imports to use exports instead


yodasw16:
@RCopeland did you coordinate with @bobwise on the promises stuff? Just making sure we aren't duplicating work.

packages/spark-core/spark-core-polyfills.js
```diff
@@ -0,0 +1,7 @@
+// Polyfills
```
afebbraro: Nice!

RCopeland:
@yodasw16 @bobwise Yep, bob's test app is to confirm the issue, this is the fix, assuming its a polyfill issue.

PR #1138, opened 2019-02-18
adding missing packages

13protons:
## What does this PR do?
Some packages were missing when cloning the repo for the first time. I added them!

### Associated Issue 
No issue associated with this (that I know of! It's my first day!)


13protons:
@RCopeland I added the missing packages w/ `npm install` and that got my test suite passing. Can you please double check the the auto populated version numbers for these packages are actually the ones we want?

afebbraro:
Glad to see you making PRs Alan! So awesome! For the peer deps, a few are required by the eslint-lint create react app package which we decided today we no longer need. We’re using the Airbnb style guide for limiting instead. So we don’t need those peerDeps.

yodasw16:
@13protons did you see @afebbraro's comments above? Does this just need to be closed or is there some follow-up that needs to happen?

13protons:
> @13protons did you see @afebbraro's comments above? Does this just need to be closed or is there some follow-up that needs to happen?

Sorry, I missed this. I opened the PR because I couldn't install and get spark running without manually installing these. I think this means that our users are in the same boat and if that's the case then these peer deps should either be called out clearly in our docs or simply included in order to reduce on-boarding friction.

13protons:
Also I'm not the expert here on version numbers etc, so any other necessary changes should just be made without my input

afebbraro:
This is a pretty important issue so I made an official git issue for this, so that we can fully look into what all we need and not. https://github.com/sparkdesignsystem/spark-design-system/issues/1160. Closing this PR so the whole issue can be addressed. Thanks!

Issue #1178, opened 2019-03-13
Masthead inbound links fail to hide masthead on mobile


13protons:
## Describe The Bug
Clicking inbound links, aka `/home` in the Angular masthead do not close the masthead component on narrow layouts. This will cause the masthead to remain in view and requires that a user manual close the masthead. 

The only indication that navigation has occurred is link highlighting in the masthead and the full url in the address bar (which may be hidden while browsing on mobile)

## To Reproduce
1. Add an Angular spark masthead component with an inbound link (like `/home`)
1. View this component on a narrow screen (aka, mobile, but dev tools is ok)
1. Click the masthead to reveal the link list
1. Click on a lick on a link in that list
1. Notice that the masthead menu stays expanded and must be manually dismissed

## Expected Behavior
Inbound links should close the masthead menu when they result in a page content change

## Desktop (please complete the following information):
Chrome, windows, simulating mobile with dev tools

Issue #1182, opened 2019-03-13
Duplicate props in Masthead docs


13protons:
## Describe The Bug

The `narrowLink` field in the props table is duplicated under the [Masthead Default](https://sparkdesignsystem.com/patterns/components/masthead#masthead-default) variant section. 

## To Reproduce
1. Go to [https://sparkdesignsystem.com/patterns/components/masthead#masthead-default](https://sparkdesignsystem.com/patterns/components/masthead#masthead-default)
2. Scroll to the props table
3. Check it out: <img width="1021" alt="Duplicate props in docs" src="https://user-images.githubusercontent.com/1456258/54313993-e0d31000-45b0-11e9-9b37-b3cb4f9de726.png">
]()

## Expected Behavior
Only one row describing `narrowLinks`

## Screenshots
🙄

## Desktop (please complete the following information):
 - OS: macOS 10.14.2
 - Browser: Chrome 72

## Additional Context
Found while researching #1178 and https://github.com/sparkdesignsystem/spark-design-system/releases/tag/untagged-236e3e692ad9315b75c0

PR #1183, opened 2019-03-14
Fix bug #1178 & #1182

corcorankt1:
## What does this PR do?
Closes the mobile navigation when router link navigation is complete.
Removes duplicate prop in Angular Docs

### Associated Issue 
issue #1178 
issue #1182 

then please remove it.

### Documentation

### Code

### Accessibility

  - [ ] Google Chrome (Mobile)
  - [ ] Apple Safari (Mobile)


13protons:
@corcorankt1 I looked at the deploy preview site and the masthead doesn't seem to have any links in it that would allow us to test this behavior. Am I missing something or does this need to be added?

corcorankt1:
@13protons I had to replace the whole navigation on the docs site locally to get a single instance of the masthead loading. We can't commit this change, so that's the only way to test is right now. Amber was sitting with me when we confirmed it works. I can sit with you and re-create if you'd like.

PR #1190, opened 2019-03-20
update links

afebbraro:
## What does this PR do?
Update code for angular links.

### DIFF
--- a/src/angular/projects/spark-core-angular/src/lib/components/sprk-link/sprk-link.component.ts
+++ b/src/angular/projects/spark-core-angular/src/lib/components/sprk-link/sprk-link.component.ts
@@ -30,7 +30,7 @@ export class SparkLinkComponent implements OnInit {
   @Input()
   linkType: string;
   @Input()
-  href = '#';
+  href: string;
   @Input()
   idString: string;
   @Input()
@@ -54,6 +54,8 @@ export class SparkLinkComponent implements OnInit {
   @Input()
   target: string;
   @Input()
+  fragment: string;
+  @Input()
   additionalClasses: string;
   @Input()
   isDisabled: boolean;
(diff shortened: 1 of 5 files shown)


src/angular/projects/spark-core-angular/src/lib/components/sprk-link/sprk-link.component.ts
```diff
@@ -61,36 +61,63 @@ export class SparkLinkComponent implements OnInit {
   isExternal = false;
 
   ngOnInit() {
+    // Sets the default href if none provided
```
13protons: Can we get a test where `href="someUndefinedProp"` to make sure angular doesn't cast `undefined` to "`undefined`", as that would break the parser here. It might not matter as `/undefined` is a valid url even it it breaks things. Probably not a deal breaker as TS is only at build time not runtime

src/angular/projects/spark-core-angular/src/lib/components/sprk-link/sprk-link.component.ts
```diff
@@ -61,36 +61,63 @@ export class SparkLinkComponent implements OnInit {
...
     return value === '#';
   }
 
+  getPathWithoutHash(value): string {
+    return value.split('#')[0];
+  }
+
+  scrollToId() {
+    const elementID = this.href.split('#').pop();
+    const element: HTMLElement = document.getElementById(elementID);
```
13protons: what about `<a name="something">`?

http://www.tagindex.net/html/link/a_name.html

src/angular/projects/spark-core-angular/src/lib/components/sprk-link/sprk-link.component.ts
afebbraro: That's a good idea. Working on that now!

src/angular/projects/spark-core-angular/src/lib/components/sprk-link/sprk-link.component.ts
yodasw16: Name is obsolete in HTML5

src/angular/projects/spark-core-angular/src/lib/components/sprk-link/sprk-link.component.ts
afebbraro: It works as expected for when the value of undefined gets passed in and it allows 'undefined' to be passed in as the href value. Which, as you mentioned is what we want. So, I think we are good because we have to allow the route of /undefined. Added a test!

Issue #1119, opened 2019-02-11
[Bug] Angular Inputs: Date Picker Input Calendar does not appear on click


GMCtree:
## Describe The Bug
When using the date picker component in a non-Dev environment (test/beta/prod), the date picker calendar does not appear as expected. This allows a user to manually type in a date, which does not trigger the max-date validation. We have also noticed that the icon does not appear in the input field to denote it is a date picker.

## To Reproduce
1. Go to the date picker component
2. Click inside of the input box
3. Date Picker calendar **should not** appear
4. Type in a date that is greater than the current date, specifically where the month is greater than the current month
5. MaxDate validation won't trigger and this will allow the future date to be accepted as valid input

## Expected Behavior
We expect the date picker calendar to appear when we click the input field, and we expect the validation to occur even if the user types in a value manually

## Screenshots
Incorrect validation after no calendar picker appears
![microsoftteams-image 6](https://user-images.githubusercontent.com/10014685/52562888-eaa8fe00-2dcd-11e9-8282-1fa472734d4d.png)

Typing in a future date does not trigger the validation
![microsoftteams-image 7](https://user-images.githubusercontent.com/10014685/52562953-16c47f00-2dce-11e9-870f-2cf8d3d9f9a2.png)

## Desktop (please complete the following information):
 - OS: Windows 10 Enterprise
 - Browser: Chrome
 - Version: 71

## Smartphone (please complete the following information):
 - Device: N/A
 - OS: N/A
 - Browser: N/A
 - Version: N/A

## Additional Context
N/A

PR #1282, opened 2019-04-09
[Bug] Angular Inputs: Date Picker Input Calendar does not appear on click

RVMendoza:
## What does this PR do?
When angular is built in --prod, angular date picker input does not work. This is because it's not being packaged correctly with the flag.

Date picker uses a third party library called `tiny-date-picker`. There are a [two ways](https://nitayneeman.com/posts/how-to-add-third-party-library-in-angular-cli/) to bring in external libraries: **Injection**, and **import modules**.

In order to use injection, the library has to have typings, which `tiny-date-picker` does not have.

**Instead of creating one, we are opting to use the second method of importing it for the sake of simplicity.**

#### Concerns addressed:

- There were concerns from @RCopeland that the import module method might an Angular anti-pattern.
  - I couldn't find sources that say it is. Injection vs import is more of a choice based on situation, but neither have been considered bad practice
  - [More conversations on the pros and cons of each](https://softwareengineering.stackexchange.com/questions/336767/module-requiring-vs-dependency-injection-in-javascript)
- Injection would be useful if we needed access to the class constructor of components and services
  - We don't need that for tiny-date-picker

> At a high level, Angular reads the NgModule metadata during the bootstrap process and creates an injector capable of providing these dependencies through the class constructor of components and services throughout the application. -- [angularfirst.com](https://angularfirst.com/when-to-use-es2015-modules-instead-of-angular-dependency-injection-and-when-not-to/)

Tiny Date picker doesn’t need any of these things.

- Injection makes that library global
  - We don't need that for `tiny-date-picker`

Further reading:
- Here's [an opinion ](https://medium.com/@bluepnume/dependency-injection-in-angular-isn-t-worth-it-more-lessons-learned-from-scaling-paypal-checkout-2189ec9c21a0)that thinks injection should be completely out of the question
- https://stackoverflow.com/questions/43861935/angularjs-dependency-injection-vs-es6-import

### Associated Issue 
#1119

then please remove it.

### Code
 - [x] Build Component in Spark Vanilla (Sass, HTML, JS)
 - [x] Build Component in Spark React

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)

### DIFF
--- a/src/angular/projects/spark-core-angular/src/lib/directives/inputs/sprk-datepicker/sprk-datepicker.directive.ts
+++ b/src/angular/projects/spark-core-angular/src/lib/directives/inputs/sprk-datepicker/sprk-datepicker.directive.ts
@@ -1,13 +1,11 @@
 import { Directive, ElementRef, Inject, Input, OnInit } from '@angular/core';
+import TinyDatePicker from 'tiny-date-picker';
 
 @Directive({
   selector: '[sprkDatepicker]'
 })
 export class SprkDatepickerDirective implements OnInit {
-  constructor(
-    public ref: ElementRef,
-    @Inject('TinyDatePicker') public TinyDatePicker: any
-  ) {}
+  constructor(public ref: ElementRef) {}
 
   @Input()
   sprkDatePickerConfig: object;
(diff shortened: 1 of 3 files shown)


RVMendoza:
fixes #1119

Issue #1243, opened 2019-04-03
[Doc] Fix Spark React code copy blocks in docs


afebbraro:
**User Story:**
As a user when I copy code from the React examples in the doc site I want it to be properly formatted so that I don't have to spend time cleaning up the formatting before it can work.

**Notes:**
- Currently: When you copy code from the react examples in the doc site it is poorly formatted. There's extra spaces in closing tags which break react apps. Would be ideal to be able to copy code and paste it into an app and not have to spend time fixing the formatting before it can work.
- **IDEA:** Have 2 preview code sample in our docs 1 - has js and 1- the data. might want a bigger fix when we design the doc site. (Rob suggestion) 
- **IDEA:** https://prismjs.com/plugins/normalize-whitespace/

**AC:**
- When I copy code from a React example in the doc site it will not be poorly formatted.
- I will not have to spend time clearing up the formatting after a copy paste.

PR #1300, opened 2019-04-16
[Doc] Fix Spark React code copy blocks in docs

RVMendoza:
## Before
copyable code blocks have extra spaces that break the code
Screen Shot 2019-04-16 at 1 16 16 PM

## After
jsx support is added to Beautifier which doesn't cause the extra spaces
image

## Additional Improvements
Added a config that puts attributes into separated lines. Makes everything more readable and match closer to how we as front-end devs would implement a component. (Affects react, angular and vanilla)

### Before
image

### After
image

### Associated Issue 
#1243

### DIFF
--- a/gulp/drizzle/config.js
+++ b/gulp/drizzle/config.js
@@ -150,6 +150,8 @@ module.exports = {
       indent_size: 2,
       indent_with_tabs: false,
       max_preserve_newlines: 1,
+      e4x: true,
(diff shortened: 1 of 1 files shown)


RVMendoza:
Fixes #1243

afebbraro:
THANK YOU FIXING THIS!! @RVMendoza ! That always annoyed the heck outta me.

PR #1357, opened 2019-04-26
release notes update

afebbraro:
## What does this PR do?
Updates release notes page and adds a redirect for the developers page that is removed.


_redirects
```diff
@@ -1 +1,2 @@
 https://fall.sparkdesignsystem.com/* https://staging.sparkdesignsystem.com/:splat 301!
+https://sparkdesignsystem.com/gettingstarted/developers https://www.sparkdesignsystem.com/ 301!
```
yodasw16: should we do the same for the designers page that we deleted?

afebbraro: Ope! 😂

PR #1558, opened 2019-07-05
setup jest config, fixes imports in a few test files, set up coverage

RCopeland:
cd spark/manifests/spark-react
    npm test


spark/components/modals/react/SprkModal.test.js
```diff
@@ -2,7 +2,6 @@ import React from 'react';
 import Enzyme, { mount, shallow } from 'enzyme';
 import sinon from 'sinon';
 import Adapter from 'enzyme-adapter-react-16';
-import '../windowStubs';
```
afebbraro: @RCopeland looks like you removed an import to windowStubs here. Was that intentional?

RCopeland: i put it into the setup file directly, no need for a separate file.

PR #1559, opened 2019-07-05
add storybook build to build command, update componentPage

RCopeland:


react/src/templates/ComponentPage.js
```diff
@@ -19,7 +19,7 @@ function ComponentPage({ data }) {
         <ContentIframe
           title="Preview"
           variant=""
-          src={`/iframe.html?id=components-${node.name}`}
+          src={`/iframe.html?id=components-${node.name}--default`}
```
afebbraro: @RCopeland, Wondering what this change is for? We know that adding in '--default' will not work for all components and their variants because some have story variants that aren't called 'default'

RCopeland: I added it so that a couple components would work, just trying to get something building on netlify.

Issue #1500, opened 2019-07-02
[Build] Floating Label for HUGE Input - Vanilla  


ssabo1:
**User Story:**
As Spark, we want to have a floating label for Sparks Huge Input so that the placeholder text as well as the label are both with in the Huge Input.

**Notes:**
- This is for Vanilla
- This Request is coming from Brendon 
- This is coming from the RMA UI Kit
- VALUES NEED TO MATCH CURRENT SPARK  
- this is a varient of Default 

**Design Q&As:**
Q: Animation requirements? Brendan working with Alan
Q: Do we still want a version without the label? YES
Q: Monetary/Percentage, Label is optional? visually Yes. but should always be in the back for accessibility. 

**AC:**
- I will see floating Labels on Spark Huge Input. 
- Update docs in reference to the label display or hide.
- BASE, I will see Label/Placeholder text Align Left, inside Huge Input. (Values need to Match current Spark.) (Fill: #FFFFFF 100% Boarder: 1px inside, #CFCFCF Radius: 4px)
- On FOCUS, Field will have Highlighted State. 
- On FOCUS, Text will Align left. 
- On FOCUS as cursor is placed inside input. Label will display inside Huge Input, top left corner.
- On FOCUS, Padding: Align left with 24px padding from left of the Input to the Text. 8px from the top of the Input to the Label and 12px from the placeholder text/text to the bottom of the input.
- User will be able to type into input.
- On ERROR validation, Input will have a highlighted state
- On ERROR validation, Error Messaging will display 12px below HUGE input. (Fill: #FFFFFF 100% Boarder: 2px inside, #EDA302 Radius: 4px)  
- CURRENCY, will display placeholder text, align left.
- CURRENCY, Visual label is optional to the display but label will always be in the background for a screen reader
- CURRENCY, as cursor is placed inside input. Label will display inside Huge Input, top left corner.

![image](https://user-images.githubusercontent.com/46761151/60536185-b5953d80-9cd3-11e9-84c4-00b5b41fa6d2.png)

13protons:
Here's a CSS only mockup to build upon: 

https://codepen.io/alanguir/pen/mZLXjN

This pen has known issues in Edge. See this compatibility chart: https://caniuse.com/#feat=css-placeholder-shown

Issue #1120, opened 2019-02-11
[BUILD] Design Tokens


yodasw16:
Here's some research to look through.

Articles
- https://medium.com/eightshapes-llc/tokens-in-design-systems-25dd82d58421
- https://medium.com/@kylegach/the-themes-are-always-changing-but-the-tokens-stay-the-same-258d57b18f31
- https://uxdesign.cc/design-tokens-for-dummies-8acebf010d71

Systems
- https://www.lightningdesignsystem.com/design-tokens/
- https://github.com/Shopify/polaris-tokens
- https://comet.discoveryeducation.com/visual-language/tokens.html
- https://github.com/rei/rei-cedar-tokens

Tools
- https://github.com/salesforce-ux/theo
- https://github.com/amzn/style-dictionary
- https://github.com/NateBaldwinDesign/dragoman
- https://github.com/mrmartineau/design-system-utils

PR #1589, opened 2019-07-15
Floating label for huge input

guern1kn:
## What does this PR do?
Adds floating label functionality to huge input

### Associated Issue 
#1500 

### Documentation
 - [x] Update Component Sass Var/Class Modifier table

### Code
 - [x] Build Component in Spark Vanilla (Sass, HTML, JS)

### Accessibility

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)

### Design Review
 - [x] Design reviewed and approved


RVMendoza:
@afebbraro fixing the direct selector feedback in the next commit

RVMendoza:
@afebbraro I agree with you about keeping our downward spacing flow.

I believe this code was following what already existed right above it. Probably why that choice was made.
image

I put a ToDo on the code when we start working on icon on huge input later (we lessened scope so we can move forward)

packages/spark/settings/_settings.scss
```diff
@@ -443,15 +443,17 @@ $sprk-text-input-font-weight: 300 !default;
 $sprk-text-input-height: 48px !default;
```
13protons: @guern1kn I don't think you should change anything right now, but I think we need to discuss our variable naming and inheritance strategy as a team. There's so many things here that should probably just inherit form other vars and not just be their stand values.

guern1kn: Sounds like a good meta topic!

yodasw16: This is probably part of a bigger discussion about Design Tokens
https://github.com/sparkdesignsystem/spark-design-system/issues/1120

bobwise:
Observed VoiceOver behavior:
image

We may want to take another look at how the structure is being interpreted for screen readers. Huge Input behavior is a little bit different from other inputs.

Expected screen reader behavior:
1. First focus and describe the label - "Text Input Label"
2. Then focus and describe the input
 2a. read the placeholder - "Placeholder"
 2b. then read the label - "Text Input Label"
 2c. then describe the state - "edit text"

packages/spark/base/inputs/_svg-icon-input.scss
```diff
@@ -4,7 +4,15 @@
 
 .sprk-b-TextInputIconContainer .sprk-c-Icon {
   margin-left: $sprk-input-icon-offset-x;
+  // TODO: We should be using margin bottom, not margin top
```
RCopeland: do we need an issue for this?

guern1kn: Probably? I'll send a note to steph.

Issue #1534, opened 2019-07-03
Unit Test & variable% Code Coverage - Stepper React


ssabo1:
**User Story:**
As Spark, we want to get React Stepper Unit Test to 100% coverage 

**Notes:**
- Spark React Stepper had been merged for the July 1st Release, but there is some outstanding work needed to complete the component.
- There is an additional Issue for adding the Carousel functionality. #1482  

**AC:**
- Unit Test.
- React Stepper will have 100% Code Coverage

PR #1590, opened 2019-07-15
React Stepper - more unit tests

bobwise:
## What does this PR do?
Add more unit tests to React Stepper

### Associated Issue 
Issue https://github.com/sparkdesignsystem/spark-design-system/issues/1534


bobwise:
These two tests are left commented out because they execute a block of code that calls `console.warn`. We don't currently have a way to mock `console`, so there is no way to assert that the correct result has occurred in the test. These tests can be uncommented to reach 100% code coverage.

Issue #1535, opened 2019-07-03
Slider Text Animation will fade text in and out - Stepper React


ssabo1:
**User Story:**
As Spark, we want `sprk-c-Stepper__slider--active` on and off so that text will fade in and out in the slider. 

**Notes:**
- Spark React Stepper had been merged for the July 1st Release, but there is some outstanding work needed to complete the component.

**AC:**
- Slider text animation. 
- Turning `sprk-c-Stepper__slider--active` on and off will fade the text in and out within the slider.  
May be useful: https://www.npmjs.com/package/react-animate-on-change 
- all 3 frameworks will use the same technique

PR #1597, opened 2019-07-22
React Stepper slider animations

bobwise:
## What does this PR do?
Add CSS animation for React Stepper

### Associated Issue 
https://github.com/sparkdesignsystem/spark-design-system/issues/1535

### Code
 - [x] Build Component in Spark React

### Accessibility
- [ ] New changes abide by [accessibility requirements](https://sparkdesignsystem.com/docs/accessibility)

  - [X] Google Chrome - Current
  - [X] Google Chrome - 1 Prior
  - [x] Google Chrome (Mobile) - Current
  - [ ] Google Chrome (Mobile) - 1 Prior (version 74)
  - [x] Mozilla Firefox - Current
  - [x] Mozilla Firefox - 1 Prior
  - [x] Mozilla Firefox (Mobile) - Current (version 18 on iOS, version 68 Android not tested)
  - [ ] Mozilla Firefox (Mobile) - 1 Prior (version 17 on iOS and/or version 68 on Android)
  - [x] Microsoft Edge - Current
  - [x] Microsoft Edge - 1 Prior
  - [x] Apple Safari - Current
  - [x] Apple Safari - 1 Prior
  - [x] Apple Safari (Mobile) - Current
  - [x] Apple Safari (Mobile) - 1 Prior

### Design Review
 - [X] Design reviewed and approved

### DIFF
--- a/packages/spark/components/_stepper.scss
+++ b/packages/spark/components/_stepper.scss
@@ -26,11 +26,6 @@
     opacity: 0;
   }
 
-  .sprk-c-Stepper__step-heading,
-  .sprk-c-Stepper__step-description {
-    transition: opacity $sprk-stepper-slider-text-fade-in-timing ease 0.2s;
-  }
-
   .sprk-c-Stepper__step-icon {
     transition: opacity $sprk-stepper-slide-timing ease;
   }
@@ -40,6 +35,7 @@
 .sprk-c-Stepper__slider--active .sprk-c-Stepper__step-icon,
 .sprk-c-Stepper__slider--active .sprk-c-Stepper__step-description {
   opacity: 1;
+  transition: opacity $sprk-stepper-slider-text-fade-in-timing ease 0.2s;
 }
 
 .sprk-c-Stepper--has-dark-bg {
@@ -212,7 +208,6 @@
 
 .sprk-c-Stepper__step-description {
   padding-top: $sprk-stepper-step-description-top-spacing;
-  transition: opacity 0.1s ease-in-out;
 }
 
 .sprk-c-Stepper--has-carousel {
--- a/packages/spark/components/_stepper.scss
+++ b/packages/spark/components/_stepper.scss
@@ -35,7 +35,7 @@
 .sprk-c-Stepper__slider--active .sprk-c-Stepper__step-icon,
 .sprk-c-Stepper__slider--active .sprk-c-Stepper__step-description {
   opacity: 1;
-  transition: opacity $sprk-stepper-slider-text-fade-in-timing ease 0.2s;
+  transition: opacity $sprk-stepper-slider-text-fade-in-timing ease $sprk-stepper-slider-text-fade-in-delay;
 }
 
 .sprk-c-Stepper--has-dark-bg {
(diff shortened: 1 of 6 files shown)


packages/spark/components/_stepper.scss
```diff
@@ -40,6 +35,7 @@
 .sprk-c-Stepper__slider--active .sprk-c-Stepper__step-icon,
 .sprk-c-Stepper__slider--active .sprk-c-Stepper__step-description {
   opacity: 1;
+  transition: opacity $sprk-stepper-slider-text-fade-in-timing ease 0.2s;
```
13protons: @afebbraro do you see any issue with changing this transition to *only* apply to the active slider? The point here is to fade in, but not out. If you look at the way vanilla operates, the text and title of the stepper only ever **fade-in**, and so this change to the css technically does not change anything about how that interaction is rendered. If anything there's now just fewer unintended side effects.

src/react/projects/spark-react/src/SprkStepper/components/SprkStepperSlider/SprkStepperSlider.js
```diff
@@ -1,8 +1,31 @@
...
 class SprkStepperSlider extends Component {
+  constructor(props) {
+    super(props);
+
+    this.sliderRef = React.createRef();
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.sliderStyle.top != this.props.sliderStyle.top){
+      this.sliderRef.current.classList.remove('sprk-c-Stepper__slider--active');
```
13protons: can this cause/effect relationship be unit tested? Aka, change `top`, observe `--active` flicker? Not sure if that adds value now, but would stop it from changing in the future.

packages/spark/components/_stepper.scss
afebbraro: Hmm, I actually didn't write the sprk-c-Stepper__slider--active class name and I think I had it on the selected step originally. But, scoping the transition in this way seems fine to do and like you said doesn't change anything about how it works currently. So, if it solves some weird React bug then I don't see harm in doing this, especially like you said, if it adds more predicability. Let's do it.

afebbraro: Oh! But can you make the delay a var? @bobwise. The hard-coded 0.2s should be placed as var in our settings file.

src/patterns/components/stepper/collection.yaml
```diff
@@ -50,7 +53,7 @@ variableTable:
   $sprk-stepper-dark-icon-color-selected:
     default: $sprk-white
     description: The color of the step icon when the step is selected and the stepper has a dark background (sprk-c-Stepper--has-dark-bg).
-  $sprk-stepper-dark-icon-color-hoverg:
+  $sprk-stepper-dark-icon-color-hover:
```
afebbraro: Good spelling catch!

src/react/projects/spark-react/src/SprkStepper/components/SprkStepperSlider/SprkStepperSlider.js
RCopeland: i think componentWIllReceiveProps is deprecated also

afebbraro:
Approved and asked Bob to update the animations issue bug with it also being an issue in Vanilla.

PR #1651, opened 2019-07-30
1602 update side navigation component

RCopeland:


src/components/ComponentIndex.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+const ComponentIndex = ({ id, name, reactLink, angularLink, htmlLink }) => (
+  <div className="component-index" id={id}>
+    <h2 className="sprk-b-TypeDisplayThree component-index__heading">{name}</h2>
+    <ul className="component-index-list">
+      { reactLink
+      && (
+      <li className="component-index-list__item">
```
afebbraro: Could change this to use one <li> and update the parts of the URL that need to be updated.

src/components/context/APIDocumentationMenu.js
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import { StaticQuery, graphql } from 'gatsby';
+
+const APIDocumentationMenu = () => (
+  <StaticQuery
+    query={graphql`
+        {
+          allDirectory(filter: {relativePath: {regex: "/^components\\//"}, relativeDirectory: {eq: "components"}}) {
```
afebbraro: Could remove the allDirectory and use the allMdx calls instead.

PR #1691, opened 2019-08-09
Fix angular animations bug with toggles/accordions

afebbraro:
## What does this PR do?
Fix bug with animations not working correctly on Spark Angular Toggles and Accordions in certain browsers.

### Associated Issue 
Fixes (Issue Number that this PR is closing out)

then please remove it.

### Documentation

### Code

### Accessibility

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)


13protons:
@afebbraro is there a strong reason to apply these styles the "angular animation way" vs. just doing whatever we do in React or Vanilla?

afebbraro:
@13protons , Good question, the reason is because the vanilla component uses a DOM manipulation library and for React library. Using the built in ng animations gives us access to a special Angular setting of  "height: *" for this instance. This solves our use case without having to incorporate a third party lib or reach into the DOM ourselves. We can also use the BrowserAnimations API to do more complex features in the future. Personally,  and I know this is a hot take haha but I prefer the ng animations over the 3rd party libs we have in vanilla and React.

13protons:
> @13protons , Good question, the reason is because the vanilla component uses a DOM manipulation library and we use a 3rd partly lib for React (a React specific lib). Using the built in ng animations gives us access to a special Angular setting of "height: *" for this instance. This solves our use case without having to incorporate a third party lib or reach into the DOM ourselves. We can also use the BrowserAnimations API to do more complex features in the future. Personally, and I know this is a hot take haha but I prefer the ng animations over the 3rd party libs we have in vanilla and React.

Perhaps we set a future story to investigate no libraries for any framework? Are we just trying to solve the jumpiness when going from `height: SOMEpx` to `height: auto` with those libraries?

PR #1709, opened 2019-08-16
Bump lodash version to 4.17.15

bobwise:
## What does this PR do?
Update the dependency version for the react site, react components, and angular components. The Angular PeerDependency has been changed to a Dependency.

### Associated Issue 
Fixes https://github.com/sparkdesignsystem/spark-design-system/issues/1719.

### Code

### DIFF
--- a/src/angular/projects/spark-angular/package.json
+++ b/src/angular/projects/spark-angular/package.json
@@ -10,16 +10,16 @@
     "lerna-publish": "cd ../../dist/spark-angular && npm publish"
   },
   "peerDependencies": {
-    "@angular/animations": ">=7.0.0 <=8.0.0",
-    "@angular/common": ">=7.0.0 <=8.0.0",
-    "@angular/compiler": ">=7.0.0 <=8.0.0",
-    "@angular/compiler-cli": ">=7.0.0 <=8.0.0",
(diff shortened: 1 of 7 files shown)


src/angular/projects/spark-angular/package.json
```diff
@@ -29,6 +29,7 @@
     "zone.js": "^0.8.26"
   },
   "dependencies": {
-    "lory.js": "^2.5.3"
+    "lory.js": "^2.5.3",
+    "lodash": "^4.17.15"
```
afebbraro: Did you mean to leave lodash as a devDependency and also add it as a dependency? Do we need both?

bobwise: lodash was added as a dependency and removed as a peerDependency. It should only appear once in this file.

bobwise: Does anything need to be changed here or can we resolve the review?

afebbraro: @bobwise, looks like you are changing the angular and spark versions to an old version here. Can you remove those changes? Them being '>=7.0.0 <=8.0.0' has been changed and the spark version has been updated too.

PR #1763, opened 2019-09-12
Angular Masthead Enhanced Scrolling Functionality

### LINKED ISSUE #1640: [Build] Percentage Input - Angular
**User Story:** 
As a Spark User, I want the ability to use a Percentage Input so that input allows for the user to calculate a percentage  

**Notes:**
Requestor: Janina, Brand Marketing 
Was a topic during Design Sync on 8/29
Icon can be found in #1743 

**AC:**
- The Percentage Input is used for 'entering a percentage amount'
- Spark recommends that this component be used when ...
- Under Inputs I will see a Percentage Input - Angular
- Percent icon will display to the right of the input
- Input will NOT extend to the current width of inputs, input will be built with 5 numeric characters in mind, build using a width that looks good and confirm with Tommy. (Communicate with engineers who are building Vanilla #1639 and React #1641 )
- Input will have selected State (ex. Monetary)

### LINKED ISSUE #1768: Masthead position:sticky not working in Safari - Angular
**User Story:**
Safari doesn't like position:sticky so masthead does not work like all other implementations.

**Notes:**
additional information: https://stackoverflow.com/questions/51792783/safari-positionsticky-not-working-in-an-overflowauto-element
- this is for angular versions on ios
- apparently ios prevents scroll events on purpose
- https://github.com/sroze/ngInfiniteScroll/issues/104
- https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Using_Touch_Events

**AC:**
- Safari will play nice with position:sticky

RVMendoza:
## What does this PR do?
Enhances the scrolling functionality of the masthead in Angular. When the user scrolls down the page on a narrow viewport, the masthead animates off the page to create more space for content. When the user scrolls back up towards the top of the page, the masthead animates back in to give easy access to the navigation.
64639640-3ceae580-d3d6-11e9-8d46-dcbe57b5a679

#### Improvements needed in the future:
- Router event: Instead of using this router, make run this on every menu link instead
- `.sprk-c-Masthead--hidden` modifier: apply it only on breakpoints so we can reduce polling for resize events (more detail in comments of masthead scss)

### Associated Issue 
Fixes #1640, Fixes #1768

then please remove it.

### Documentation

### Code
 - [x] Unit Testing in Spark Angular with `gulp test-angular` (~100%~ coverage, 100% passing)

#### Line 240 -246 of `sprk-masthead.component.ts`
- Skipping testing because the function inside of it is being tested already, and there's no need to test angular to see if it's doing it's job.
- this will be replaced by improvements mentioned above.

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Internet Explorer 11 (only this specific version of IE) -- expected functionality for it not to stick
  - [x] Microsoft Edge
  - [x] Apple Safari - found an issue where sticky doesn't work correctly
  - [x] Apple Safari (Mobile)

### DIFF
--- a/src/angular/projects/spark-angular/src/lib/components/sprk-masthead/sprk-masthead.component.spec.ts
+++ b/src/angular/projects/spark-angular/src/lib/components/sprk-masthead/sprk-masthead.component.spec.ts
@@ -39,6 +39,7 @@ describe('SprkMastheadComponent', () => {
     component.scrollDirection = 'up';
     component.isHidden = false;
     component.isNarrowLayout = false;
+    component.isNarrowOnResize = false;
     window.document.body.style.minHeight = '800px';
     window.document.body.style.minWidth = '1024px';
   });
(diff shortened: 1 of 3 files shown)


packages/spark/components/_masthead.scss
```diff
@@ -37,6 +37,17 @@
   transform: $sprk-masthead-translateY;
   transition: $sprk-masthead-transition;
 }
+// TODO: Improve --hidden modifier
```
RCopeland: i think you should remove the comment block and put it in an issue instead

guern1kn: @RVMendoza What are your thoughts? I kind of agree that this should be in the issue.

RVMendoza: took it out, will add it to issue

PR #1892, opened 2019-09-19
Percent Vanilla Updates

guern1kn:
## What does this PR do?
Update class table and variable names for the Vanilla Percent Input


packages/spark/settings/_settings.scss
```diff
@@ -623,12 +623,10 @@ $sprk-input-text-icon-offset-x: $sprk-space-m !default;
 $sprk-input-text-icon-offset-y: 2em !default;
 $sprk-input-text-icon-font-size: 1rem !default;
 $sprk-input-text-icon-font-weight: 700 !default;
-$sprk-text-input-has-text-icon-right-height: 1em !default;
-$sprk-text-input-has-text-icon-right-width: 1em !default;
```
guern1kn: The height and width are being inherited from .sprk-c-Icon, so these are no longer needed.

PR #2074, opened 2019-09-25
Clarifying documentation on masthead component.

erikwoods:
After speaking with @afebbraro, she confirmed that `$mobile-masthead-height` was a typo, and that the unit of measurement should be in pixels.

### DIFF
--- a/src/patterns/components/masthead/default.hbs
+++ b/src/patterns/components/masthead/default.hbs
@@ -12,7 +12,7 @@ information:
 restrictions:
 - You must set the variable $sprk-masthead-narrow-height so that the navigation displays properly on small screens.
   You should put your logo in the Masthead and then grab the total height of the Masthead with your logo in it. Set
-  $mobile-masthead-height to that height.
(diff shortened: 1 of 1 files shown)


src/patterns/components/masthead/default.hbs
```diff
@@ -12,7 +12,7 @@ information:
 restrictions:
 - You must set the variable $sprk-masthead-narrow-height so that the navigation displays properly on small screens.
   You should put your logo in the Masthead and then grab the total height of the Masthead with your logo in it. Set
-  $mobile-masthead-height to that height.
+  $sprk-masthead-narrow-height to that height, in pixels (e.g. $sprk-masthead-narrow-height: 55px).
```
bobwise: Hi Erik. Thank you for contributing to Spark! 

The `:` in this line is breaking our build. Can you add another commit with this character escaped?

```suggestion
  $sprk-masthead-narrow-height to that height, in pixels (e.g. $sprk-masthead-narrow-height":" 55px).
```

Issue #2096, opened 2019-10-03
[BUG] Huge Text Input with Default Value in spark-react not moving label on page load


RCopeland:
## Describe The Bug
Huge Text Input with Default Value in spark-react not moving label on page load spark-react in edge/ie11/firefox

## To Reproduce
1. Go to [https://deploy-preview-2094--sparkdesignsystem-react.netlify.com](https://deploy-preview-2094--sparkdesignsystem-react.netlify.com)
2. Click on Inputs, then Text Inputs
3. Scroll down to 'Text Huge with Default Value'

#Screenshots
<img width="565" alt="Capture" src="https://user-images.githubusercontent.com/412903/66147672-4c2c1000-e5dd-11e9-8c3b-f9cf743c54c5.PNG">

PR #2099, opened 2019-10-03
Fix Huge Text Input Default Value bug

bobwise:
In IE11, huge text inputs with a default value should render correctly on page load.

https://github.com/sparkdesignsystem/spark-design-system/issues/2096

### DIFF
--- a/src/react/projects/spark-react/src/SprkInput/components/SprkTextareaCheck.js
+++ b/src/react/projects/spark-react/src/SprkInput/components/SprkTextareaCheck.js
@@ -9,6 +9,22 @@ class SprkTextareaCheck extends Component {
     this.state = {
       hasValue: this.props.value,
     };
+
+    this.inputRef = React.createRef();
+  }
+
+  componentDidMount() {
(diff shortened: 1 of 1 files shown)


src/react/projects/spark-react/src/SprkInput/components/SprkTextareaCheck.js
```diff
@@ -11,6 +11,16 @@ class SprkTextareaCheck extends Component {
...
+  componentDidMount() {
+    if (this.props.type === 'hugeTextInput') {
+      if (this.props.value !== '') {
+        this.setState({
+          hasValue: true,
+        });
+      }
+    }
+  }
+
```
RCopeland: i dont like that SprkTextareaCheck is the owner of this state. I think it'd be better as part of SprkTextInput.

bobwise: Thanks Rob. Why does it make more sense to have this in SprkTextInput?

PR #2138, opened 2019-10-07
Storybook React Alerts - Knobs and Actions

bobwise:
## What does this PR do?
Add Knobs and Actions for React Alerts in Storybook.

### Associated Issue 
Fixes https://github.com/sparkdesignsystem/spark-design-system/issues/1896 (knobs)
and Fixes https://github.com/sparkdesignsystem/spark-design-system/issues/1895 (actions).

### DIFF
--- a/spark/components/alerts/react/SprkAlert.stories.js
+++ b/spark/components/alerts/react/SprkAlert.stories.js
@@ -15,7 +15,6 @@ export const defaultStory = () => (
   <SprkAlert
     onFocus={action('focus')}
     message={text('Message', 'This is an alert.')}
-    additionalClasses={text('additionalClasses', '') || undefined}
     iconNameDismiss={text('iconNameDismiss', '') || undefined}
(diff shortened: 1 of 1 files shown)


bobwise:
My reasoning for what should be a knob and what should be an action:

# Which Props should be Knobs?

## Yes
`message` - yes, but needs to be fixed so it doesn't allow HTML
`isVisible`
`isDismissible`
`iconNameDismiss`

`iconNameFail` - yes, but only for the Fail story.
`iconNameInfo` - yes, but only for the Info story
`iconNameSuccess` - yes, but only for the Success story

## No
`variant` - these should be stories
`idString` - no visual change
`analyticsString` - no visual change
`additionalClasses` - no value - only works for our utility classes

# Actions

## Yes
`onDismiss` - triggered with onClick of Dismiss Button/icon

## No
`onFocus` - not part of our API

Issue #2149, opened 2019-10-11
fix failing tests for react pagination component in NDS


afebbraro:
Pagination needs the failing tests fixed.

```
Summary of all failing tests
 FAIL  components/pagination/react/SprkPagination.test.js (5.279s)
  ● renders ellipses with more than 3 pages

    expect(received).toBe(expected) // Object.is equality

    Expected: 2
    Received: 4

       97 |   expect(paginationItems.find('a.sprk-c-Pagination__icon').length).toBe(2); // two chevrons
       98 |   expect(paginationItems.find('a.sprk-c-Pagination__link').length).toBe(3); // three pages
    >  99 |   expect(paginationItems.findWhere(x => x.text() === '...').length).toBe(2); // two ellipses
          |                                                                     ^
      100 | });
      101 | 
      102 | it('navigates back one page when you click the first chevron', () => {

      at Object.toBe (components/pagination/react/SprkPagination.test.js:99:69)

  ● goes to hard-coded page 1 when you click 1

    Method “simulate” is meant to be run on 1 node. 2 found instead.

      143 |     />,
      144 |   );
    > 145 |   wrapper.find('a.sprk-c-Pagination__link').findWhere(x => x.text() === '1').simulate('click');
          |                                                                              ^
      146 |   // first param of the first call
      147 |   expect(spyFunc.mock.calls[0][0].newPage).toBe(1); // expect it to have "1" as arg.newPage
      148 | });

      at ReactWrapper.single (manifests/spark-react/node_modules/enzyme/build/ReactWrapper.js:1718:17)
      at ReactWrapper.simulate (manifests/spark-react/node_modules/enzyme/build/ReactWrapper.js:952:21)
      at Object.simulate (components/pagination/react/SprkPagination.test.js:145:78)

  ● goes to hard-coded page 2 when you click 2

    Method “simulate” is meant to be run on 1 node. 2 found instead.

      157 |     />,
      158 |   );
    > 159 |   wrapper.find('a.sprk-c-Pagination__link').findWhere(x => x.text() === '2').simulate('click');
          |                                                                              ^
      160 |   // first param of the first call
      161 |   expect(spyFunc.mock.calls[0][0].newPage).toBe(2); // expect it to have "2" as arg.newPage
      162 | });

      at ReactWrapper.single (manifests/spark-react/node_modules/enzyme/build/ReactWrapper.js:1718:17)
[...]

PR #2151, opened 2019-10-14
Fix React Pagination Unit Tests

bobwise:
## What does this PR do?
Make test selectors more specific so tests pass.  It is unclear why these tests started failing when we switched to Storybook.

### Associated Issue 
Fixes https://github.com/sparkdesignsystem/spark-design-system/issues/2149

---

This line is returning different results depending on which site we run the tests in.

`expect(paginationItems.findWhere(x => x.text() === '...').length).toBe(2);`

Old Site - finds two nodes:
image

New Site - finds 4 nodes:
image


afebbraro:
I think being more specific in the tests is a great idea and we should merge this but I am really curious why this worked before and it didn't now. Tagging @RCopeland  for his thoughts before we merged this. Rob, any ideas as to how this worked in the Old structure?

PR #2152, opened 2019-10-14
Storybook vanilla alert

bobwise:
## What does this PR do?
Set up stories and initial renders for vanilla Alert components

### Associated Issue 
Fixes https://github.com/sparkdesignsystem/spark-design-system/issues/1793

### DIFF
--- a/spark/components/alerts/vanilla/alerts.stories.js
+++ b/spark/components/alerts/vanilla/alerts.stories.js
@@ -7,7 +7,7 @@ export default {
 export const info = () => {
   useEffect(() => {
     alerts();
-  });
+  }, []);
 
   return `
     <div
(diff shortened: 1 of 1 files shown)


spark/components/alerts/vanilla/alerts.stories.js
```diff
@@ -7,7 +7,7 @@ export default {
 export const info = () => {
   useEffect(() => {
     alerts();
-  }, []);
+  });
```
afebbraro: Hey @bobwise why did you remove the [] I was passing as a second argument into useEffect here?

bobwise: what does this array do?

afebbraro: It makes it so the alerts function only runs once and that our effect doesn’t depend on any values from props or state, so it never needs to re-run. @bobwise

bobwise: ahh ok I found it in the docs. I didn't realize the empty array was a special use case. I'll put it back.

PR #2365, opened 2019-11-19
breaks out the inputs into separate stories (React)

### LINKED ISSUE #2357: [Research] - Figure out what to do about the inputs stories. We might need to break them up so that we get propstables for each one. Especially an issue because we don’t have a component per spark input either. Some of them have only “container” components.

RCopeland:
## What does this PR do?
Breaks out inputs into separate stories in react sb.

### Associated Issue 
Fixes #2357 

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)


afebbraro (changes requested):
Nice, this seems to be exactly what we wanted for breaking them out. For the structure, I think we should change the story names from what you have. For our other stories we have the Name of the component like "Link" and then it toggles open the stories which are the variants. So Instead of having a toggle that says "Text Input" that shows a story called "Text Input" we would have a toggle that says "Text Input" that shows a story called "Default". I was thinking that we could possibly move the "Huge Select" and "Huge Text" stories into their respective base folders too. So, inside the "Select" toggle, you would see two different stories. One would be "Default" and the other would be "Huge".

afebbraro (approved):

PR #2434, opened 2019-12-19
add sassdoc comments for link && input

### LINKED ISSUE #2416: Add comments to the scss files for the class modifiers - inputs & links
**User Story:**
As Spark we want to add comments to the SCSS Files for Class Modifiers.

**Notes:**
Requestor: Katie after building #2374 

**AC:**
- add comments to the scss files for INPUT class modifiers. 
- add comments to the scss files for LINK class modifiers.

guern1kn:
## What does this PR do?
Adds comments to Link and Input for sassdocs.

### Associated Issue 
Fixes #2416 

Created new issue for Input Table to render #2487 

### Documentation
 - [x] Update Component Sass Var/Class Modifier table

### Screenshots
image


html/base/_input_comments.scss
```diff
@@ -0,0 +1,27 @@
+////
```
guern1kn: I do not like this solution, but there is an issue with Sassdoc not parsing the subdirectories, so it isn't going through the Input directory to grab these.

afebbraro: hey @guern1kn ! Could you try crafting a different regex for Sassdoc so it parses the subdirectories?

guern1kn: That's the issue with Sassdocs. It doesn't parse the subdirectories 2 levels in regardless of the regex.

afebbraro: Did you see this logged in an issue on Sassdocs? I want to take a look at their filed issue if there is one. @guern1kn

guern1kn: There is not one specifically for this yet. I read as part of them building the parser that it was difficult to build the ability to parse through sub directories. I will try and find that again.

afebbraro: Okay, let me know, i'm marking this as changes requested till then. I do not want to let limitations in tooling dictate how we structure our code. Especially, since this problem will only ever get larger in the future. I don't believe this is the right step. I would like to see more research done on the sassdocs side. Whether that is pulling up an old issue, reaching out to them for support on how we can achieve what we want or having myself or the whole team jump in to help you research this.

guern1kn: That's fine by me. I'll separate out the Link portion of this to get it merged and get with Steph about creating a separate research issue for the Input task.

guern1kn: @afebbraro What I'm going to do is get the comments in the desired spot for both link and inputs, then get an issue made with Steph to research the Sassdocs parsing the input directory. That way at least the comments are already there when we resolve the parsing problem.

Issue #2419, opened 2019-12-18
Add 2 Red Color variables to Spark 


ssabo1:
**User Story:**
As Spark, we want to add two Red color variables to Spark so that we align with Brand Central for consistency across the board. 

**NOTES:**
Requestor: Jane Ugrinovskiy 12/18/19 When looking at Brand Central Jane realized two colors are missing from Spark, Jane talked to Jenn Hillard and Jenn agreed these would be good to include. 
- Complete with #2176 

**AC:**
- I will see Deep red #76232F added to Spark
- I will see Mid red #9D2235 added to Spark

Issue #2372, opened 2019-11-21
[Build] New Page Design - Color 


ssabo1:
Work with Tommy for a better layout.

**User Story:**
As Spark we want enhance the Design of our Color doc in Gatsby so that it updated on the doc site.

**Notes:**
Work with Tommy for a better layout.

**AC:**
- write css to display colors.
- styling the page to copy and past code blocks easily

ssabo1:
Worked with Jane to create New Color Page Layout for Doc Site MVP.

PR #2483, opened 2019-12-23
add auto-generated Color page to Gatsby site!

afebbraro:
## What does this PR do?
Add Color page to Gatsby site and two new shades of red.

There will be a follow-up task to finish the A11y and Dos/donts section.

### Associated Issue 
#2372 and #2419 

### Code
 - [x] Build Component in Spark Gatsby Site

### Accessibility
- [ ] New changes abide by [accessibility requirements](https://sparkdesignsystem.com/docs/accessibility)
---- Will be working on this with olya once back from ooo

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)


afebbraro:
This is ready for review but will have additional work done for the Do's / Don't and a11y sections. Everything but those sections can be reviewed.

src/components/ColorSwatch.js
```diff
@@ -0,0 +1,80 @@
...
+      sprk-o-Stack--large
+      "
+    >
+      <h3 className="sprk-b-TypeDisplayFive sprk-o-Stack__item">
+        {name}
+      </h3>
+
+      <p className="sprk-b-TypeBodyTwo sprk-o-Stack__item">
+        {description}
+      </p>
```
bobwise: I don't think this is providing any info that we're not already getting from the name. I think we can delete this.

bobwise: I don't think this is providing any value that isn't already in name. I think we can delete this.

src/components/Header.js
```diff
@@ -8,6 +8,7 @@ const Header = ({ menuVisible, setMenuVisible, context, setContext }) => (
   <header className="docs-header">
     <div className="docs-header__group">
       <button className="docs-header__menu-button" onClick={() => { setMenuVisible(!menuVisible); }}>
+        <span className="sprk-u-ScreenReaderText">Toggle Navigation</span>
```
bobwise: 👍

src/components/ColorSwatch.js
afebbraro: hey @bobwise thanks for the comment, the description is something that I need to nail down with Jane. I think we might use this for something like Usage. Which could be really helpful to show someone when to use a particular color. If after syncing with Jane, we decide we don't need it then I'll remove it.

afebbraro:
@bobwise  you mentioned maybe separating out the package version updates. I think I should include them here cause they're needed to be able to view this color page without having to wait 10 mins. I tested out the gatsby site and didn't see any issues with the new update, but if something comes we can always roll back, update again or fix it in a follow up PR.
Since we are in rapid development with this stuff, I've been updating Storybook and other deps as much as I possibly can in hopes to keep us up to date. Once we have a stable build that's ready for prod (when we get all our component content into gatbsy) then I think we should absolutely be more mindful about doing version updates. (We talked IRL about this!) Woot!!

yodasw16:
@afebbraro I missed the time to submit a review, but I think it needs more space between the paragraph at the top and the images. I would do `mbh`. On small view ports the white background of the image creates extra space that isn't there when the images are side-by-side. Maybe put a light grey border around the white image or a box shadow like the cards have so that you can see the edges?

PR #2521, opened 2020-01-06
quoting glob

RCopeland:
## What does this PR do?
Allows sassdocmodifiers to parse sub folders


guern1kn:
I actually had to add this in all of the package.json files for netlify to build each storybook instance correctly, so it needs to be updated in all 4.

guern1kn (approved):

afebbraro (approved):

PR #2638, opened 2020-01-17
update markdownDocumentationLinkBuilder throughout

### LINKED ISSUE #2542: Update the 'markdownDocumentationLinkBuilder' to accept a 'componentType' parameter
**User Story:**
As spark we want to update the 'markdownDocumentationLinkBuilder' to accept a 'componentType' parameter so that it will allow us to use it in the typography and spacing stories. 

**Notes:**
Requestor: Katie 1/8/20 when adding type documentation to NDS

**AC:**
- Update the 'markdownDocumentationLinkBuilder' to accept a 'componentType' parameter 
- Update the typography and spacing stories.

guern1kn:
## What does this PR do?
Updates the markdownDocumentationLinkBuilder to accept an additional param. 
Updates all stories with new param.

### Associated Issue 
Fixes #2542 

### Documentation
 - [x] Update Spark Docs React

  - [x] Google Chrome
  - [ ] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [ ] Microsoft Edge
  - [x] Apple Safari
  - [ ] Apple Safari (Mobile)


storybook-utilities/markdownDocumentationLinkBuilder.js
```diff
@@ -1,6 +1,6 @@
-const markdownDocumentationLinkBuilder = (component) => {
+const markdownDocumentationLinkBuilder = (type, component) => {
```
afebbraro: I would make the 'type' default to 'components' to save having to supply 'components' in almost all the story files. We only have 1-3 cases where the type is not going to be components.

storybook-utilities/markdownDocumentationLinkBuilder.js
```diff
@@ -1,6 +1,7 @@
-const markdownDocumentationLinkBuilder = (type, component) => {
+const markdownDocumentationLinkBuilder = (component, type) => {
```
afebbraro: Hey @guern1kn, thanks for making the suggested change but you could do with by setting a default parameter. ES6 allows you to define a default for a param if its undefined. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters.  

```suggestion
const markdownDocumentationLinkBuilder = (component, type='components') => {
```
So you don't need the whole line `const directory = !type ? 'components' : type;`.

Doing it as part of the argument definition makes it clearer what the func is doing right when you read the available options.

Issue #2366, opened 2019-11-19
[Add] Content For How to Install Spark Fonts - NDS


ssabo1:
**User Story:**
As Spark, we created 'How to Install Spark Fonts Doc' in word that we want to copy into Gatsby so that our New Doc Site displays the updated documentation.

**Notes:**
- Docs can be found at shorty/sparkcontent  

**AC:**
- I will see updated documentation displaying on the New Doc Site

PR #2658, opened 2020-01-22
Add Installing Fonts Guide to NDS

bobwise:
## What does this PR do?
Adds the "Getting Started with Spark Fonts" guide to the New Docs Site and updates some layout stuff in the Icons guide.

### Associated Issue 
Fixes https://github.com/sparkdesignsystem/spark-design-system/issues/2366

  - [X] Google Chrome
  - [ ] Google Chrome (Mobile)
  - [ ] Mozilla Firefox
  - [ ] Microsoft Edge
  - [ ] Apple Safari
  - [ ] Apple Safari (Mobile)


src/components/markdown-render/blockquote.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+function Blockquote(props) {
```
afebbraro: Lets use es6 arrow functions here.

src/components/markdown-render/table.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+function Table(props) {
```
afebbraro: lets use es6 arrow funcs here too

src/scss/components/_blockquote.scss
```diff
@@ -0,0 +1,6 @@
+.docs-b-Blockquote {
+  background-color: $sprk-alert-bg-color-info;
```
afebbraro: I love this blockquote style with that lighter blue! It's cute! Can you update this to use the blue color itself and not the color from the alert, just in case we change the alert color in Spark. Don't want to then have that accidentally change our Blockquotes in our NDS.

src/components/markdown-render/blockquote.js
afebbraro: lets use es6 arrow function here.

src/components/markdown-render/table.js
afebbraro: Lets use es6 arrow func here too.

src/scss/components/_blockquote.scss
bobwise: yes good catch! I meant to change this but forgot.

PR #2719, opened 2020-02-05
version bumps for publish

### LINKED ISSUE #2679: Publish Spark Packages
- [ ] publish spark packages
- [ ] update spark react in Gatsby and be sure everything works

afebbraro:
## What does this PR do?
 Update version numbers in packages.

### Associated Issue
Fixes #2679 

then please remove it.

### Code
 - [x] update version numbers in packages

### DIFF
--- a/react/.npmignore
+++ b/react/.npmignore
@@ -1,3 +1,4 @@
+node_modules/
 coverage/
 .storybook/
 storybook-static/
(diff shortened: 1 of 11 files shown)


react/.npmignore
```diff
@@ -1,3 +1,3 @@
-node_modules/
```
RCopeland: i think this needs to be there, so that we don't package node_modules with the publish

afebbraro: I noticed we don’t have it in master and we don’t have it for vanilla or angular. Should we add it?

PR #2731, opened 2020-02-06
add export for SprkStepper

guern1kn:
## What does this PR do?
Add export for SprkStepper.


afebbraro:
we just need the SprkStepperStep in addition to what you have

PR #2775, opened 2020-02-12
change htmlFor to id on input

### LINKED ISSUE #2195: [Bug] TextareaCheck sets htmlFor={id} when it should set id={id} ... Spark inputs dont have Id attrbute
## Describe The Bug
Currently spark ID attributes don't have ID. Mistake in prop setting causes a for attribute to get rendered on an input / textarea instead of the id

**Notes:**
- came from a talk with Patrick
- TextareCheck has been renamed to spark input element
- Currently broke today.

**AC:**
- All Spark inputs will have and ID attribute in React
- TextareaCheck will set id={id}

RVMendoza:
## What does this PR do?
Change htmlFor attribute on input to id

### Associated Issue
Fixes #2195

### Documentation
 - [x] Update Spark Docs

### Code

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)

### DIFF
--- a/react/src/base/inputs/SprkErrorContainer/SprkErrorContainer.js
+++ b/react/src/base/inputs/SprkErrorContainer/SprkErrorContainer.js
@@ -33,7 +33,7 @@ SprkErrorContainer.propTypes = {
    */
   analyticsString: PropTypes.string,
   /**
-   * Configured by parent and assigned to the `htmlFor` attribute.
+   * Assigned to the `id` attribute Automatically configured if not provided.
    */
   id: PropTypes.string.isRequired,
   /**
--- a/react/src/base/inputs/SprkErrorContainer/SprkErrorContainer.js
+++ b/react/src/base/inputs/SprkErrorContainer/SprkErrorContainer.js
@@ -33,7 +33,7 @@ SprkErrorContainer.propTypes = {
    */
   analyticsString: PropTypes.string,
   /**
-   * Assigned to the `id` attribute Automatically configured if not provided.
+   * Configured by parent and assigned to the `id` attribute.
    */
   id: PropTypes.string.isRequired,
   /**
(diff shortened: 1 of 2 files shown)


react/src/base/inputs/components/SprkInputElement/SprkInputElement.js
```diff
@@ -67,7 +67,7 @@ class SprkInputElement extends Component {
           'sprk-b-InputContainer__input--has-icon-right': iconRight,
         })}
         type={type}
-        htmlFor={id}
+        id={id}
```
guern1kn: Instead of replacing the `htmlFor` can we add the `id={id}` while maintaining support for `htmlFor` in case someone is using it? It would mean both would render with the `id` but I think removing it and changing the API qualifies as a breaking change.

RVMendoza: htmlFor is only for "for" and should be put on labels. It wasn't working or affecting anything, and this is correcting it because inputs should have IDs, not `for`

```
<label htmlFor="username">Username</label>
<input name="username" id="username" type="text"/>
```

guern1kn: Ok. Looks like the propType description for id needs to be updated. References htmlFor.

RVMendoza: I'm seeing the user isn't able to set id on this in general. Adding that functionality now

RVMendoza: changed!

react/src/base/inputs/SprkErrorContainer/SprkErrorContainer.js
```diff
@@ -33,7 +33,7 @@ SprkErrorContainer.propTypes = {
    */
   analyticsString: PropTypes.string,
   /**
-   * Configured by parent and assigned to the `htmlFor` attribute.
+   * Assigned to the `id` attribute Automatically configured if not provided.
```
guern1kn: Need a period in here after attribute

Issue #2830, opened 2020-03-05
[Enhance] Will see ID or explicit button that allows user to link to specific component variations. - All Packages


ssabo1:
**User Story:**
As a user I want to see an ID or explicit button so that I can to link to specific component variations.

Current state: there are no ID or explicit button to allow user to do it.

**Notes:**
Requestor: Alan on 2/28, This is a super common interaction on github readme files and documentation in general.
- This is for the Doc Site
 

**AC:**
- ALL PACKAGES: Heading on all Component Pages will be able to link to a specific component variation. 
- Users would like this to work: https://sparkdesignsystem.com/using-spark/components/alert#fail-alert

![image.png](https://images.zenhubusercontent.com/5c98da1df75c1c000158e457/440063ca-d22b-4ae9-ad9b-6eb7f874d145)

PR #2883, opened 2020-03-09
Add anchor id tags to headings

RVMendoza:
## What does this PR do?
All headings (h1, h2, h3, h4) wil have an id corresponding to the headline.

There is an offset because of masthead and it would scroll it to wrong part of the screen. 
an invisible anchor tag is added above the headline with the corresponding h3 so it positions correctly.

### Associated Issue
#2830 

### Documentation
 - [x] Update Spark Docs

### Accessibility

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)


src/components/markdown-render/h1.js
```diff
@@ -1,8 +1,24 @@
 import React from 'react';
 
 function H1(props) {
+  const { children } = props;
+  const id = children.toString().replace(/\s+/g, '-').toLowerCase();
   return (
-    <h1 className="docs-b-h1 sprk-b-PageTitle sprk-b-TypeDisplayOne sprk-u-mbh sprk-u-Measure" {...props}></h1>
+    <>
+      <a id={id} className="docs-b-Link--anchor" />
```
afebbraro: Hey @RVMendoza,  For jump links, you usually have an anchor with an href ex. `href="#accordion"` and then when you click that, it would go to another element (doesnt need to be an anchor) that has the id="accordion". If we plan to have another anchor that when clicked would go to these headings, then this would not be an anchor tag.

RVMendoza: oh right! I guess this could be a span just the same!

src/components/markdown-render/h1.js
```diff
@@ -1,8 +1,24 @@
 import React from 'react';
 
 function H1(props) {
+  const { children } = props;
+  const id = children.toString().replace(/\s+/g, '-').toLowerCase();
   return (
-    <h1 className="docs-b-h1 sprk-b-PageTitle sprk-b-TypeDisplayOne sprk-u-mbh sprk-u-Measure" {...props}></h1>
+    <>
+      <span id={id} className="docs-b-Link--anchor" />
```
afebbraro: This span thats wrapping the h1 tag needs a closing tag at the end of the heading tag.

afebbraro: Resolved this, the intention it to have it NOT wrap the heading.

PR #2911, opened 2020-03-13
Refactor flag into flexbox, add reverse

### LINKED ISSUE #2907: Re-write the Flag object to use Flexbox instead of table
The Flag Object uses `display: table` to allow the content to the right of the media to be vertically positioned. This should be re-written to use Flexbox instead.

This will also resolve https://github.com/sparkdesignsystem/spark-design-system/issues/1139

**AC:**
- HTML Flag object will use flexbox

RVMendoza:
## What does this PR do?
HTML - Refactor flag to use flexbox, and add reverse story
flag

### Associated Issue
Fixes #2907 

### Documentation
 - [x] Update Spark Docs

### Code
 - [x] Build Component in HTML

### Accessibility

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)

### DIFF
--- a/html/objects/_flag.scss
+++ b/html/objects/_flag.scss
@@ -8,7 +8,6 @@
 
 .sprk-o-Flag {
   display: flex;
-  align-items: flex-start;
 }
 
 .sprk-o-Flag__figure {
@@ -16,7 +15,9 @@
 }
 
 .sprk-o-Flag__body {
-  flex: 1;
+  flex-grow: 1;
+  flex-shrink: 1;
+  flex-basis: 0;
 
   > :last-child,
   & {
@@ -118,7 +119,7 @@
 
   /// Stacked flag at smaller screen width.
   .sprk-o-Flag--stacked {
-    flex-flow: column;
+    flex-direction: column;
 
     &.sprk-o-Flag--rev {
       > .sprk-o-Flag__figure {
--- a/html/objects/_flag.scss
+++ b/html/objects/_flag.scss
@@ -122,12 +122,6 @@
   .sprk-o-Flag--stacked {
     flex-direction: column;
 
-    &.sprk-o-Flag--rev {
-      > .sprk-o-Flag__figure {
-        order: 0;
-      }
-    }
-
     > .sprk-o-Flag__figure {
       padding-right: 0;
       padding-left: 0;
--- a/html/objects/_flag.scss
+++ b/html/objects/_flag.scss
@@ -28,8 +28,9 @@
 /// Reverses the flag and places the
 /// image on the right of the body content.
 .sprk-o-Flag--rev {
+  flex-direction: row-reverse;
+
   > .sprk-o-Flag__figure {
-    order: 1;
     padding-right: 0;
     padding-left: $sprk-flag-gutter;
   }
--- a/html/objects/_flag.scss
+++ b/html/objects/_flag.scss
@@ -7,27 +7,16 @@
 ////
 
 .sprk-o-Flag {
-  display: table;
-  width: 100%;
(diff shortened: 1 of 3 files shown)


html/objects/_flag.scss
```diff
@@ -7,45 +7,28 @@
...
-  display: table;
-  width: 100%;
-}
-
-.sprk-o-Flag__body,
-.sprk-o-Flag__figure {
-  display: table-cell;
-  vertical-align: top;
+  display: flex;
+  align-items: flex-start;
```
afebbraro: Why is the flex-start needed here?

html/objects/_flag.scss
```diff
@@ -7,45 +7,28 @@
...
-
-  > img {
-    display: block;
-    max-width: none;
-  }
 }
 
 .sprk-o-Flag__body {
-  width: 100%;
+  flex: 1;
```
afebbraro: flex: 1, the shorthand, has issues in ie11 sometimes. Can you take a look there.

html/objects/_flag.scss
```diff
@@ -7,45 +7,28 @@
...
 .sprk-o-Flag--rev {
-  direction: rtl;
-
-  > .sprk-o-Flag__body,
-  > .sprk-o-Flag__figure {
-    direction: ltr;
-  }
-
   > .sprk-o-Flag__figure {
+    order: 1;
```
afebbraro: What were your thoughts on using the alignment properties vs. using order?

html/objects/_flag.scss
```diff
@@ -132,15 +115,15 @@
 }
 
 @media all and (max-width: $sprk-flag-stacked-breakpoint) {
-  
+
   /// Stacked flag at smaller screen width.
   .sprk-o-Flag--stacked {
-    direction: ltr;
+    flex-flow: column;
```
afebbraro: If we are only setting column, we dont need the whole shorthand `flex-flow`. Can just just `flex-direction`

RVMendoza: You're right!

html/objects/_flag.scss
RVMendoza: I changed it to flex-direction reverse-row -- it feels easier to understand

html/objects/_flag.scss
RVMendoza: you're right, it's not necessary, it's default is flex stretch if not provided and that would be fine for this.

src/pages/using-spark/components/flag.mdx
```diff
@@ -38,6 +38,13 @@ area can be set to any Spacing value.
 - On small viewports, the content area
 can be configured to move below the media.
 
+### Reverse Modifier
```
afebbraro: Gonna message you about adding something in here about a11y with reversing the order of the elements.

RVMendoza: image

html/objects/_flag.scss
```diff
@@ -132,15 +117,15 @@
...
-    direction: ltr;
+    flex-direction: column;
 
-    > .sprk-o-Flag__body,
-    > .sprk-o-Flag__figure,
-    & {
-      display: block;
+    &.sprk-o-Flag--rev {
+      > .sprk-o-Flag__figure {
+        order: 0;
```
afebbraro: hey @RVMendoza , Did you forget to remove this order property? Looks like it's not doing anything.

Issue #2957, opened 2020-03-18
[Enhance] Masthead Component narrow nav menu with aria-controls & Expanded - HTML


ssabo1:
**User Story:**

**Notes:**
Requestor: RM-12 in the a11y audit.
SME: Bob

**AC:**
masthead narrow nav / Hamburger Menu
- Masthead Component aria-expanded will update when the menu is opened.
- Masthead Component will have aria-controls.

aria-expanded: this attribute is on the trigger element (the button). Its value is set to "true" when the element is expanded and is set to "false" when the element is collapsed.

aria-controls this attribute is on the trigger element (the button). It's value is set to the id of the content element. If the user provides an id use that. If the user doesn't provide an id, generate one with lodash.

```<button aria-controls="123"></button>
<div class="im_the_content" id="123">The content</div>

PR #2961, opened 2020-03-24
HTML Masthead - add aria-controls and aria-expanded

bobwise:
## What does this PR do?
- adds `aria-controls` functionality to the mobile navigation menu in the HTML Spark Masthead
- adds `aria-expanded` functionality to the mobile navigation menu in the HTML Spark Masthead
- adds new HTML utility functions for `toggleAriaExpanded` and `generateAriaControls` with unit tests

### Associated Issue
Fixes https://github.com/sparkdesignsystem/spark-design-system/issues/2957

### Code
 - [X] Build Component in HTML
 - [X] Unit Testing in HTML with `npm run test` in `html/` (100% coverage, 100% passing)

### Accessibility
- [X] New changes abide by [accessibility requirements](https://sparkdesignsystem.com/docs/accessibility)

  - [X] Google Chrome
  - [ ] Google Chrome (Mobile)
  - [ ] Mozilla Firefox
  - [ ] Microsoft Edge
  - [ ] Apple Safari
  - [ ] Apple Safari (Mobile)

### DIFF
--- a/html/components/masthead.js
+++ b/html/components/masthead.js
@@ -1,11 +1,12 @@
 /* global document window */
-import { throttle } from 'lodash';
+import { throttle, uniqueId } from 'lodash';
 import getElements from '../utilities/getElements';
 import { focusFirstEl } from '../utilities/elementState';
 import { isEscPressed } from '../utilities/keypress';
 import isElementVisible from '../utilities/isElementVisible';
 import scrollYDirection from '../utilities/scrollYDirection';
 import { hideDropDown, showDropDown } from './dropdown';
+import { toggleAriaExpanded } from './toggle';
 
 const addClassOnScroll = (element, scrollPos, scrollPoint, classToToggle) => {
   // If user scrolls past the scrollPoint then add class
@@ -82,6 +83,8 @@ const toggleMobileNav = (iconContainer, nav, masthead) => {
     .querySelector('svg')
     .classList.toggle('sprk-c-Menu__icon--open');
   nav.classList.toggle('sprk-u-Display--none');
+
+  toggleAriaExpanded(iconContainer);
 };
 
 const focusTrap = (isOpen, nav) => {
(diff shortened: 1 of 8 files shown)


html/components/masthead.js
```diff
@@ -1,11 +1,12 @@
 /* global document window */
-import { throttle } from 'lodash';
+import { throttle, uniqueId } from 'lodash';
 import getElements from '../utilities/getElements';
 import { focusFirstEl } from '../utilities/elementState';
 import { isEscPressed } from '../utilities/keypress';
 import isElementVisible from '../utilities/isElementVisible';
 import scrollYDirection from '../utilities/scrollYDirection';
 import { hideDropDown, showDropDown } from './dropdown';
+import { toggleAriaExpanded } from './toggle';
```
afebbraro: Hey Bob, we don’t want to create a dependency on toggle here. Can we make a util for toggleAriaExpanded ?

html/components/masthead.js
```diff
@@ -130,6 +133,38 @@ const bindUIEvents = () => {
...
+      }
+    }
+
+    // if aria-controls doesn't exist
+    // or it does exist but the ids don't match
+      // init aria-controls
+    if (!element.hasAttribute('aria-controls')
+      || element.getAttribute('aria-controls') != nav.getAttribute('id')) {
+
+      var contentId;
```
afebbraro: Should be `let`
```

File: html/components/masthead.js [hunk 3]
```diff
@@ -130,6 +133,38 @@ const bindUIEvents = () => {
...
+      // if we already have an ID, use that
+      if (nav.hasAttribute('id')){
```
afebbraro: #space the curly and paren

html/tests/masthead.tests.js
```diff
@@ -12,15 +16,113 @@ import {
...
+
+    // Add iconContainerDiv to mastheadDiv
+    mastheadDiv.appendChild(iconContainerDiv);
+    main.appendChild(mastheadDiv);
+    document.body.appendChild(main);
+  })
+
   afterEach(() => {
-    document.querySelectorAll.restore();
+    if (document.querySelectorAll.restore) {
```
afebbraro: What’s the reason for  this new if ?

bobwise: The `.restore` is cleanup from the sinon spy on line 69. Before my task, that test was the only test in that section. When I added my new tests, the `.restore` line started breaking because I wasn't creating a sinon spy that needed to be cleaned up.

html/components/masthead.js
bobwise: Let's talk about this more. I agree it should be a utility, but it's exported in spark-exports already.

afebbraro: @bobwise, if someone wants to use that utility they shouldn't have to import the toggle file. Even though toggle may be the only place it is used in Spark right now, it's important to remember that we don't code for the past, but for the future. And we have a lot of new components on the horizon. We are also currently auditing everything we have in Spark from an a11y standpoint and who knows what will come up! Whats the hesitation from adding this as a utility?

html/tests/masthead.tests.js
afebbraro: Thank you! That makes sense!

html/components/masthead.js
bobwise: talked in person. We're going to pull this functionality out into an html utility file that can be shared between the HTML toggle components. Eventually we should deprecate `toggleAriaExpanded` in toggle.js and export the new utility function in spark-exports.

bobwise: Created the `toggleAriaExpanded` and `generateAriaControls` utilities. `toggleAriaExpanded` is exported as `toggleAriaExpandedAttribute` because we already have a utility with that name.

afebbraro:
@bobwise what does the coverage report say on masthead after your changes? Before it was 97.48% Statements 116/11991.89% Branches 34/3796.55% Functions 28/2997.48% Lines 116/119

PR #3008, opened 2020-04-02
update react toggle to button

### LINKED ISSUE #2945: [Enhance] Toggle's link will be a <button> - React
**User Story:**
As Spark we want Toggle to use a button to open and close so that we are accessible. 

**Notes:**
Current State: Toggle uses a link to open and close it, since it doesn't navigate, it should be a button. 
Requestor: A11y audit, Section RM-12.
SME: Bob

**AC:**
React Toggle
- will see `<a>` replaced with `<button>` and fix the CSS.
- will see aria-controls. This can be added to the component with a lodash call if the ID doesn't exist.

aria-controls this attribute is on the trigger element (the button). It's value is set to the id of the content element. If the user provides an id use that. If the user doesn't provide an id, generate one with lodash.

```<button aria-controls="123"></button>
<div class="im_the_content" id="123">The content</div>

### PR DESCRIPTION
**Pull request by guern1kn**
## What does this PR do?
- [x] Update the React Toggle component to use <button> instead of <a>
- [x] Add aria-controls to the React Toggle component
- [x] Update tests for React Toggle component

### Associated Issue
Fixes #2945 

### Code

### Accessibility

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)

### Screenshots
image

### DIFF
--- a/react/src/components/footer/components/SprkFooterAwards/SprkFooterAwards.test.js
+++ b/react/src/components/footer/components/SprkFooterAwards/SprkFooterAwards.test.js
@@ -58,7 +58,8 @@ describe('SprkFooterAwards:', () => {
     };
 
     const wrapper = mount(<SprkFooterAwards awards={awards} />);
-    expect(wrapper.find('a').length).toBe(3);
+    expect(wrapper.find('a').length).toBe(2);
+    expect(wrapper.find('button').length).toBe(1);
   });
 
   it('should render the href as nogo if not defined', () => {
--- a/react/src/components/toggle/SprkToggle.js
+++ b/react/src/components/toggle/SprkToggle.js
@@ -34,11 +34,10 @@ class SprkToggle extends Component {
       titleAddClasses,
       iconAddClasses,
       toggleIconName,
+      ariaControls,
       ...other
     } = this.props;
     const { isOpen, height } = this.state;
-
-    const ariaControls = uniqueId('sprk_toggle_content_');
 
     const containerClasses = classnames(
       'sprk-c-Toggle',
@@ -83,6 +82,7 @@ class SprkToggle extends Component {
 
 SprkToggle.defaultProps = {
   toggleIconName: 'chevron-down-circle-two-color',
+  ariaControls: uniqueId('sprk_toggle_content_'),
 };
 
 SprkToggle.propTypes = {
@@ -112,6 +112,10 @@ SprkToggle.propTypes = {
   titleAddClasses: PropTypes.string,
   /** Additional classes for the toggle icon. */
   iconAddClasses: PropTypes.string,
+  /** A string that is used to set the aria-controls and associated id for the component. 
+   * The default is generated using lodash.uniqueId.
+  */
+  ariaControls: PropTypes.string,
 };
 
 export default SprkToggle;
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: react/src/components/footer/components/SprkFooterAwards/SprkFooterAwards.test.js [hunk 1]
```diff
@@ -58,7 +58,8 @@ describe('SprkFooterAwards:', () => {
     };
 
     const wrapper = mount(<SprkFooterAwards awards={awards} />);
-    expect(wrapper.find('a').length).toBe(3);
+    expect(wrapper.find('a').length).toBe(2);
+    expect(wrapper.find('button').length).toBe(1);
```
guern1kn, line 62: `SprkFooterAwards` uses `SprkToggle` and this test broke when updating.

afebbraro: @guern1kn do we need an issue to update SprkFooterAwards component too or is it all set?

react/src/components/toggle/SprkToggle.js
```diff
@@ -99,6 +112,10 @@ SprkToggle.propTypes = {
   titleAddClasses: PropTypes.string,
   /** Additional classes for the toggle icon. */
   iconAddClasses: PropTypes.string,
+  /** A string that is used to set the aria-controls and associated id for the component. 
+   * The default is generated using lodash.uniqueId.
```
afebbraro, line 116: We don't need to mention that we are using lodash specifically to generate the ID. All the user needs to know is that there is a default generated if they don't provide one.

react/src/components/toggle/SprkToggle.test.js
```diff
@@ -30,25 +30,25 @@ describe('SprkToggle:', () => {
...
   });
 
   it('should add aria-expanded="true" when the toggle is open', () => {
     const wrapper = mount(
       <SprkToggle title="Toggle title">Body text</SprkToggle>,
     );
-    wrapper.find('a').simulate('click');
-    expect(wrapper.find('[aria-expanded="true"]').length).toBe(2);
+    wrapper.find('button').simulate('click');
+    expect(wrapper.find('[aria-expanded="true"]').length).toBe(1);
```
afebbraro, line 52: We added the aria-controls logic so we need to also add tests for that.

react/src/components/toggle/SprkToggle.js
guern1kn: That was something I added after I saw that the documentation shows the default value as `uniqueId('sprk_toggle_content_')` but I can certainly take it back out.

react/src/components/footer/components/SprkFooterAwards/SprkFooterAwards.test.js
guern1kn: `SprkFooterAwards` utilizes `SprkToggle` so it will be updated when we finish this task. No need for a separate issue.

afebbraro: Thats good new!

react/src/components/toggle/SprkToggle.js
afebbraro: oh hmm. I still think that they probably dont care how the ID is generated or with what. They'll just assume thats a function we use.

Issue #3080, opened 2019-04-02
[New] Add Skip Nav to each Spark pages


ssabo1:
**User Story:**
As a Spark User I want to see a Skip Nav added to Spark so that my screen readers can bypass the many links before the main content and I can go to directly to what i am looking for.

**Notes:**
- This came about form an email 'Site Accessibility link on QL.com' with Joe Lapata.
- Brand Sites have added a link to the footer
 - RM Audit also called this out, No skip link provided RM-9. A skip link allows keyboard-only users, including screen reader users, to quickly move from the start of the page to the main content.
- Future Spark will add a components to design system.

**AC:**
- I will see Skip Nav added to each Spark page

PR #3103, opened 2020-04-24
Add Skip Nav to Gatsby Site

bobwise:
## What does this PR do?
1. Adds a skip nav to the top of the site that moves focus to the top of the content. This is useful for navigating the site with a keyboard or screen reader.
2. Whenever a route changes, put focus on the skip nav.
3. Update Gatsby to version 2.20.x
- This adds an `aria-live` region that announces each route change.
- Removes `role="group"` that was causing VoiceOver to sometimes read the full contents of a page on route change.

### Associated Issue
Fixes https://github.com/sparkdesignsystem/spark-design-system/issues/3080

### Documentation
 - [X] Update Spark Docs

  - [X] Google Chrome
  - [x] Google Chrome (Mobile)
  - [X] Mozilla Firefox
  - [x] Microsoft Edge
  - [X] Apple Safari
  - [x] Apple Safari (Mobile)

### Screenshots
image


afebbraro:
I think we left off on Friday with you going to try out the screenreader util classes. Let me know when this is ready for review again.

src/scss/_spark-theme.scss
```diff
@@ -49,4 +49,8 @@ body {
   visibility: hidden;
 }
 
+.docs-b-SkipNav {
```
afebbraro, line 52:
```suggestion
.docs-c-SkipNav {
```
Would be "c" for component instead of "b" for base element.

src/components/layouts/Layout.js
```diff
@@ -64,6 +64,14 @@ const Layout = ({ children, initialContext, hasSideBar, location }) => {
     `}
       render={data => (
         <>
+          <div className="sprk-o-Box docs-c-SkipNav">
+            <a
```
afebbraro, line 68: Why not use the sprk link component here instead of an a tag with the spark link classname added to it?

src/scss/_spark-theme.scss
```diff
@@ -49,4 +49,22 @@ body {
   visibility: hidden;
 }
 
+.docs-c-SkipNav {
+  position: absolute;
+  top: 0;
+  z-index: 5000;
+
+  a {
```
afebbraro, line 57: We try not to directly select HTML elements. Could you update this? Maybe docs-c-SkipNav__link ?

PR #3143, opened 2020-05-11
Add Angular CLI Schematics support to spark-angular

jtneal:
## What does this PR do?
Adds Angular schematics support. 
### Associated Issue
Fixes https://github.com/sparkdesignsystem/Backlog/issues/544

then please remove it.

### Documentation
 - [ ] Update Spark Docs

### Design Review
 - [ ] Design reviewed and approved

### To-Do
- [x] Figure out if we should update documentation.
- [x] Figure out if we can add icons or fonts automatically.
- [x] Documentation on how to develop and work with schematics locally (`npm link`)
- [x] Parse5 issue, I saw this locally but didn't realize there was a real problem.


afebbraro:
Chatting with Jason about this!

jtneal:
parse5 is supposed to be an optional dependency, but we have build errors without it.

https://github.com/angular/components/blob/master/src/cdk/package.json#L27

afebbraro:
Talked with the team and with @jtneal. We love this idea and in the process of code reviewing this PR for merge. Will put any code questions we have on here! Thanks again, @jtneal!

afebbraro:
hey @jtneal, When running the ng add spark command in an app with Angular 7 or Angular 8 this error happens:

`“The package that you are trying to add does not support schematics. You can try using a different version of the package or contact the package author to add ng-add support.”`

Seems to work for me in an Angular 9 app.

jtneal:
@afebbraro This is actually the message that `ng add` will say if you try to use it to install Spark today, so it's not a new error. Once this PR is merged and a new release with the schematics built is released, it should work fine - it's only an error with local testing when schematics are first being added to a project. We could probably try to re-write the schematics implementation rather than having an internal-only sub command we could build everything into a single command, and it would probably alleviate this issue, but I don't think it's necessary.

afebbraro:
@jtneal It works for me in an Angular 9 app and does the adding of the sprk-u-JavaScript class and everything. No error shown. I got everything all npm linked as well, so it should be using the linked version which has the schematics. It doesnt work in an angular 7 or 8 app though.

jtneal:
I know what you meant. I will try to get the test working in 7/8 - what I was saying is once merged, it will work fine in 7/8 because the problem has to do with how sub commands are processed independently in ng 7 and ng 8 but not in ng 9. So it's only a issue we see when testing this locally.

jtneal:
@afebbraro I added the fix. I tested with Angular 7, 8, and 9. Let me know if it works better for you now as well.

angular/projects/spark-angular/schematics/ng-add/index.ts
```diff
@@ -1,36 +1,180 @@
...
+
+    if (!projectIndexFiles.length) {
+      logger.error(`Could not find the index file for this project.`);
+      logger.info(`Please manually add the "${sparkAngularJavaScriptClass}" class to your HTML tag.`);
+      return;
+    }
+
+    projectIndexFiles.forEach((indexFilePath) => {
+      // This should go on <html> but it is easier for now to just use body
+      addBodyClass(tree, indexFilePath, sparkAngularJavaScriptClass);
```
afebbraro, line 175: Hey @jtneal , our expectation is that this class is added to HTML element.

jtneal: I know, as the comment states here it was much easier to place on body because @angular/cdk supports it. I will need to submit a PR to @angular/cdk to add support for updating the HTML tag, or else we'd have to copy a bunch of @angular/cdk code into this project to accomplish it. I am curious what use-case requires the class on the html tag where it wouldn't work on the body tag? That doesn't seem to make sense.

https://sparkdesignsystem.com/installing-spark/angular doesn't mention this as an installation step any more - so we should probably add that back to the installation documentation as well. In addition, I would expect we'd want it added the starter example: https://github.com/sparkdesignsystem/spark-starter-angular/blob/master/src/index.html

yodasw16: @jtneal I'm having trouble understanding if CDK is part of Material or if it's an agnostic library for custom components that Material just uses. We haven't really focused on Spark being compatible with other UI systems like Material. Either way, changing where that class is put causes API differences between our implementations and has the possibility of breaking hooks that teams have already created with that class on HTML. There's too much risk of a breaking change to our consumers if we just change it without first deprecating the old way and phasing it out.

I think we need to put this on hold until we can gauge the impact that changing the class from html to body would have downstream. I think this is addition is a win for us in the long run, but not something we can just change quickly. Obviously there's also the option of opening a PR against CDK to make it possible to add classes to the html instead, if that is something you have the time or desire to do.

yodasw16: @afebbraro Is that an option for now? I think we identified that our docs are missing the instruction to place the class on HTML and that we need to add that back in. For this, could we remove the code for adding the class to the body and add a note in the instructions that says you need to add the class to the HTML element if you are using vanilla js code (I think that's the usecase?)

afebbraro: Hey all! We could do that but you also need to add it for Angular. So, this wouldnt be a full setup for our users and would require them to do additional work. Which could be awesome that we are doing some of the work for people but we could end up confusing them. What do you think about that @yodasw16 ?

afebbraro: @yodasw16 What do you think about what I said up above?

yodasw16: @afebbraro I think that since none of our implementations automatically add the JS utility class (right?), leaving it out of this with explicit instructions in the README that say you still need to add it is just as user-friendly as everything else. I would propose that we drop the class from the Body, add instructions to the README, and open an issue on the Material repo asking them to add the ability to put a class on the HTML element.

afebbraro: @yodasw16 our Vanilla implementation adds it automatically. Your proposal sounds good to me though!

jtneal:
@afebbraro @yodasw16 I've updated the implementation to actually add the class to the HTML tag instead of the Body tag. I will submit a PR to @angular/cdk to move some of this logic I added into their package, if they're up for it. I will have to refactor a little bit of their code to reduce duplication, but if I have time this hack week I'll try doing it then.

angular/projects/spark-angular/schematics/ng-add/index.ts
```diff
@@ -0,0 +1,179 @@
...
+/**
+ * Adds the spark and spark-angular dependencies to package.json
+ */
+function addSparkDependencies(_options: Schema): Rule {
+  return (tree: Tree, context: SchematicContext) => {
+    // Schematics automatically adds to "dependencies" but it should be "devDependencies"
+    removePackageJsonDependency(tree, '@sparkdesignsystem/spark-angular');
+
+    const dependencies: NodeDependency[] = [
+      { name: '@sparkdesignsystem/spark', overwrite: true, type: NodeDependencyType.Dev, version: '^13.0.0' },
```
afebbraro, line 60:
```suggestion
      { name: '@sparkdesignsystem/spark', overwrite: true, type: NodeDependencyType.Dev, version: '^13.1.0' },
```
We had a release yesterday.

angular/projects/spark-angular/schematics/ng-add/index.ts
```diff
@@ -0,0 +1,179 @@
...
+      { name: '@sparkdesignsystem/spark-angular', overwrite: true, type: NodeDependencyType.Dev, version: '^10.0.0' },
```
afebbraro, line 61:
```suggestion
      { name: '@sparkdesignsystem/spark-angular', overwrite: true, type: NodeDependencyType.Dev, version: '^10.1.0' },
```

angular/projects/spark-angular/schematics/ng-add/index.ts
```diff
@@ -0,0 +1,179 @@
...
+function addSparkStyles(options: Schema): Rule {
+  return (tree: Tree, context: SchematicContext) => {
+    const workspace = getWorkspace(tree);
+    const project = getProjectFromWorkspace(workspace, options.project);
+    const styleFilePath = getProjectStyleFile(project);
+    const logger = context.logger;
+
+    if (!styleFilePath) {
+      logger.error(`Could not find the default style file for this project.`);
+      logger.info(`Please manually import the Spark stylesheet in your CSS.`);
```
afebbraro, line 126:
```suggestion
      logger.info(`Please manually import the Spark stylesheet in your Sass.`);
```

angular/projects/spark-angular/schematics/ng-add/index.ts
```diff
@@ -0,0 +1,179 @@
...
+      return;
+    }
+
+    const buffer = tree.read(styleFilePath);
+
+    if (!buffer) {
+      logger.error(`Could not read the default style file within the project (${styleFilePath})`);
+      logger.info(`Please manually import the Spark stylesheet in your CSS.`);
```
afebbraro, line 134:
```suggestion
      logger.info(`Please manually import the Spark stylesheet in your Sass.`);
```

angular/projects/spark-angular/schematics/ng-add/index.ts
```diff
@@ -0,0 +1,179 @@
...
+    addModuleImportToRootModule(tree, browserAnimationsModuleName, browserAnimationsModuleImport, project);
+
+    return tree;
+  };
+}
+
+/**
+ * Adds the Spark stylesheets to the project style file
+ */
+function addSparkStyles(options: Schema): Rule {
```
afebbraro, line 117: Lets use arrow functions

jtneal: I would personally prefer not do this. By keeping these as functions, the implementation details can be hidden lower down in the file while the technical concerns can be more easily visible. You can currently glance at the file and very quickly see that ngAdd does this:

```
export function ngAdd(options: Schema): Rule {
  return chain([
    addSparkDependencies(options),
    addSparkModule(options),
    addAnimationsModule(options),
    addSparkStyles(options),
    addSparkClass(options),
  ]);
}
```

By making this change, these lines would have to go at the very end of the file so that we aren't using the variables before their declaration. I'll make this change if you want, but just explaining why I don't like it.

afebbraro: Those familiar with JS would understand where to look to find the code they need to read and you also added a bunch of nice code commenting that helps. We should be using modern JS.

afebbraro:
Tested and good from my end. Waiting for @RCopeland  and @yodasw16  to do final testing

PR #3225, opened 2020-06-17
Gatsby Site - Axe Fixes

bobwise:
A11y improvements identified using Axe

## Homepage
- [X] Document must have one main landmark 
Changed `#main_content` div into a `<main>` and wrap content
- [x] All page content must be contained by landmarks
Added `aria-hidden` to powerlines SVG as this element is decorative.

## Installing Spark
- [x] Ensures landmarks are unique
The SprkMasthead and the Gatsby side nav are both `<nav>` elements with no additional label, so they have the exact same name when viewed in the Accessibility Tree - "Navigation". Fixed by giving the Gatsby nav an `aria-label`. Also applies to Principles and Using Spark.

## Using Spark
- [x] id attribute value must be unique
`H3` and `H4` templates were hard-coded to always have the `id` of the header element match the id of the jump link anchor. `H1` and `H2` had this line removed already, so I just removed it from h3 and h4 as well.

## Component Page
- [x] Heading levels should only increase by one
The Developer Preview box has a hard-coded `<h4>`.  Changed to `<h2>` to fit the page structure. Left the style the way it is - style should be distinct from semantics.


afebbraro:
@bobwise I love seeing stuff like this from you. Everything makes sense and looks great to me. Are you adding anything else to this PR or is it ready to go? I'm confused because you have both the do not merge and the ready for review labels on it. If ready for merge, should only have the ready for review label and remove the do not merge label so we know its safe to merge.

PR #3288, opened 2020-07-15
Guide typo

bobwise:
Fixes a typo in the React Installation guide


bobwise:
> How did you even notice that?

I was following the guide to set up a new app and when I copy/pasted that line my app crashed. 😉

PR #3525, opened 2020-10-02
Updated setValidSelect to be an arrow function

### LINKED ISSUE #3509: Update setValidSelect to be an arrow function
Update setValidSelect.js in the html/utilities/validation directory to be an arrow function.
PRs should be made against the staging branch.

joanwanjiku:
Signed-off-by: Joan_E <joanevans18@gmail.com>

## What does this PR do?
Update setValidSelect to be an arrow function

### Associated Issue
Fixes #3509


guern1kn (changes requested):
Thanks for the PR! The arrow function looks good. Let's remove the package-lock from the PR and I'll get it merged!

guern1kn (changes requested):
@joanwanjiku Thanks for the update. Sorry if my previous comment wasn't the most clear. We don't want to delete the package-lock.json, but not have any changes to it in this PR. So the only file that should have changes is the setValidSelect.js.

joanwanjiku:
> @joanwanjiku Thanks for the update. Sorry if my previous comment wasn't the most clear. We don't want to delete the package-lock.json, but not have any changes to it in this PR. So the only file that should have changes is the setValidSelect.js.

PR #3533, opened 2020-10-02
fix: Fix non-unique id on toggle and AccordionItem

### LINKED ISSUE #3472: Cannot use uniqueID on defaultProps (React: SprkToggle, SprkMastheadAccordionItem)
## Describe The Bug
If you were to put two Toggles on the same page, they would not have unique Ids, even though they are being set on defaultProps (which is the problem).

We cannot use `uniqueID()`  on `defaultProps` because `defaultProps` is a static property, which means it doesn't change on every new class.

This makes sense because if the value is different on every instance of the class, it wouldn't be considered default, and in your case, it shouldn't be.

Source: https://stackoverflow.com/questions/57628582/reactjs-difference-between-generates-a-unique-id-via-default-props-and-state

### React Components Affected:
- SprkToggle
- SprkMastheadAccordionItem

## To Reproduce
1. Go to SprkToggle story
2. Copy Toggle and make it show twice
3. Build, and observe that the `id`s and corresponding accessibility attributes are all the same 

## Expected Behavior
If multiple components exist on the same page, their ID and accessibility attribute pairing should be unique to each instance.

Kalovelo:
## What does this PR do?
The default proptype uniqueID use at
- SprkToggle
- SprkMastheadAccordionItem

did not create unique IDs. Therefore now there's a check whether if prop with unique id is passed, and if not, a unique id is generated.

Relative tests were also updated.

### Associated Issue
Fixes #3472 

### Code

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [ ] Apple Safari
  - [ ] Apple Safari (Mobile)

### Screenshots
image

### DIFF
--- a/react/src/components/masthead/components/SprkMastheadAccordionItem/SprkMastheadAccordionItem.js
+++ b/react/src/components/masthead/components/SprkMastheadAccordionItem/SprkMastheadAccordionItem.js
@@ -38,11 +38,12 @@ class SprkMastheadAccordionItem extends Component {
       leadingIcon,
       subNavLinks,
       text,
+      itemId,
       ...rest
     } = this.props;
     const { isOpen, height, subNavLinks: stateLinks } = this.state;
     const TagName = element;
-    const itemId = rest.itemId ? rest.itemId : uniqueId('sprk_accordion_item_');
+    const uniqueIdentifier = itemId || uniqueId('sprk_accordion_item_');
     return (
       <li
         className={classNames(
@@ -61,7 +62,7 @@ class SprkMastheadAccordionItem extends Component {
               className="sprk-c-MastheadAccordion__summary"
               onClick={this.toggleAccordionOpen}
               aria-expanded={isOpen ? 'true' : 'false'}
-              aria-controls={itemId}
+              aria-controls={uniqueIdentifier}
               type="button"
             >
               <span className="sprk-b-TypeBodyOne sprk-c-MastheadAccordion__heading">
@@ -78,7 +79,7 @@ class SprkMastheadAccordionItem extends Component {
             <AnimateHeight duration={300} height={height}>
               <ul
                 className="sprk-b-List sprk-b-List--bare sprk-c-MastheadAccordion__details"
-                id={itemId}
+                id={uniqueIdentifier}
               >
                 {stateLinks.map((subnavlink) => {
                   const {
--- a/react/src/components/masthead/components/SprkMastheadAccordionItem/SprkMastheadAccordionItem.js
+++ b/react/src/components/masthead/components/SprkMastheadAccordionItem/SprkMastheadAccordionItem.js
@@ -38,11 +38,11 @@ class SprkMastheadAccordionItem extends Component {
       leadingIcon,
       subNavLinks,
       text,
-      itemId,
       ...rest
     } = this.props;
     const { isOpen, height, subNavLinks: stateLinks } = this.state;
     const TagName = element;
+    const itemId = rest.itemId ? rest.itemId : uniqueId('sprk_accordion_item_');
     return (
       <li
         className={classNames(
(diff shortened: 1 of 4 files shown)


react/src/components/masthead/components/SprkMastheadAccordionItem/SprkMastheadAccordionItem.js
```diff
@@ -38,11 +38,11 @@ class SprkMastheadAccordionItem extends Component {
       leadingIcon,
       subNavLinks,
       text,
-      itemId,
       ...rest
     } = this.props;
     const { isOpen, height, subNavLinks: stateLinks } = this.state;
     const TagName = element;
+    const itemId = rest.itemId ? rest.itemId : uniqueId('sprk_accordion_item_');
```
RVMendoza, line 45: @Kalovelo Could you walk me through the advantages of turning `itemId` into a rest parameter instead of the typical way of declaring it as a prop then checking for a value?

Kalovelo: Thank you for your comments!

Since it it used 
here 
https://github.com/sparkdesignsystem/spark-design-system/blob/79e495d9bfb3cb03fd8f3ff70233bde68e95378f/react/src/components/accordions/components/SprkAccordionItem/SprkAccordionItem.js#L68
and here
https://github.com/sparkdesignsystem/spark-design-system/blob/79e495d9bfb3cb03fd8f3ff70233bde68e95378f/react/src/components/accordions/components/SprkAccordionItem/SprkAccordionItem.js#L87

Conditionally setting a unique id to those places with `itemId || uniqueId('sprk_accordion_item_')` would end up with 2 different IDs, therefore a new variable with the same checking would keep the pass the same value for each prop.

Instead of creating a new `const`, I removed the `id` deconstruction and kept the original name.

Should I declare it back to prop and create a new variable which imeplements the 'itemId || uniqueId('sprk_accordion_item_')' checking?

RVMendoza: > Should I declare it back to prop and create a new variable which imeplements the 'itemId || uniqueId('sprk_accordion_item_')' checking?

Yes please, that would better match how we're doing it in the rest of the codebase. 

I am curious if your way has some performance upgrades/ better js convention with it. Maybe we could improve the code further down the line if that's the case!

react/src/components/toggle/SprkToggle.js
```diff
@@ -35,15 +36,15 @@ class SprkToggle extends Component {
...
-      contentId,
       ...other
     } = this.props;
     const { isOpen, height } = this.state;
 
-    const containerClasses = classnames(
-      'sprk-c-Toggle',
-      additionalClasses
-    );
+    const contentId = other.contentId
```
RVMendoza, line 43: Let's make the same changes on this file based on our conversation on SparkMastheadAccordionItem

react/src/components/toggle/SprkToggle.js
```diff
@@ -105,7 +106,9 @@ SprkToggle.propTypes = {
    * Assigned to the `data-analytics` attribute serving as a unique selector for outside libraries to capture data.
   */
   analyticsString: PropTypes.string,
-  /**
+  /** Whether if the toggle is open or not */
```
RVMendoza, line 109: Let's  clarify this description by saying something like,

"Determines if the toggle is open upon loading on the page."

react/src/components/masthead/components/SprkMastheadAccordionItem/SprkMastheadAccordionItem.js
RVMendoza: We'll also need tests for these instances. Thank you for your help with this!

Kalovelo:
Thank you for your detailed feedback!

PR updates:
- setup a new variable for ID prop checking.
- I also updated the SprkToggle tests since it used to test a static ID.
- SprkMastheadAccordionItem now tests whether if the generated id is unique.

RVMendoza (approved):
This looks good to me @Kalovelo, I'll have the code owners take a look as well!

guern1kn (approved):

PR #3538, opened 2020-10-03
Fix focus for React modal component

### LINKED ISSUE #3531: React Modal does not trap focus
## Describe The Bug
The React Modal does not trap focus when the modal is open.

## To Reproduce
1. Go to 'https://html.sparkdesignsystem.com/?path=/story/components-modal--default-story'
2. Click in the modal to give it focus
3. Press Tab several times. Note that the focus indicator cycles within the modal while it is open.
4. Go to 'https://react.sparkdesignsystem.com/?path=/story/components-modal--default-story'
5. Click in the modal to give it focus
6. Press Tab several times. Note that the focus indicator is not trapped in the modal and is able to move into other parts of the UI.

## Expected Behavior
When a modal is open, focus should be "trapped" inside the control.

KaylaKremer:
## What does this PR do?
Fixes the issue of the React modal not trapping the focus to elements inside of it. The issue was that the value for boolean checks for seeing if tab/escape keys are pressed or if the element is active was not being returned. Therefore, the part of the `SprkModal.js` code that sets the focus to the modal elements was never being hit.

### Associated Issue
Fixes #3531 

### Code
 - [ x ] Build Component in React

  - [ x ] Google Chrome

### Screenshots
Screen Shot 2020-10-02 at 9 43 53 PM

### DIFF
--- a/react/src/components/modals/SprkModal.js
+++ b/react/src/components/modals/SprkModal.js
@@ -68,11 +68,11 @@ class SprkModal extends Component {
   }
 
   isTabPressed(e) {
-    e.key === 'Tab' || e.keyCode === 9;
+    return e.key === 'Tab' || e.keyCode === 9;
   }
   
   isEscPressed(e) {
-    e.key === 'Escape' || e.keyCode === 27;
+    return e.key === 'Escape' || e.keyCode === 27;
   }
 
   getFocusableEls(containerRef) {
@@ -87,7 +87,7 @@ class SprkModal extends Component {
   }
 
   isActiveElement(elementRef) {
-    document.activeElement === elementRef;
+    return document.activeElement === elementRef;
   }
 
   setExternalFocus() {
@@ -164,7 +164,7 @@ class SprkModal extends Component {
         this.cancel();
       }
     }
-
+    
     if (this.isTabPressed(e)) {
       if (variant === 'wait') {
         e.preventDefault();


bobwise:
Thanks for the PR @KaylaKremer ! I'll pull this down and get it tested and verified.

bobwise:
Hi @KaylaKremer! Thanks again for this fix; looks like you also fixed a bug with the Escape key not working that we hadn’t even noticed!

How would you feel about adding some unit tests to this PR? We’d like to have tests that verify that focus stays inside the component when pressing the Tab key and also Shift+Tab. You can see some similar tests for the HTML Modal in `modals.test.js`.

KaylaKremer:
@bobwise 

You're welcome! :)

I took a quick look at the `modals.test.js`. Looks like testing to see if focus stays in the component is already covered with `should return focus to the first element if tab is pressed while the last focusable element is focused` & `should return focus to the first element if tab + shift is pressed while the first focusable element is focused`.

Unfortunately, writing unit tests is not my strong point and I'm not very familiar with Jest. I'd try to spend more time on it, but busy with working on my company's own codebase. Might be another good Hacktoberfest issue for someone else to pick up though!

bobwise:
Thank you for your contribution @KaylaKremer! I expect this PR to be approved and merged soon!

guern1kn (approved):

PR #3548, opened 2020-10-05
React Award API Updates

### LINKED ISSUE #3315: Award Upgrades - Part 2 (React)
**Notes:**
Angular Changes (this will be done in sparkdesignsystem/spark-design-system#3314
 
**React Changes**
- The React component will render the disclaimer (renders an empty Toggle where you can see the toggle icon and click it but theres no text) if none is supplied whereas the angular component has a disclaimer input that makes that section optional. Should update React component to not render the toggle if no toggle info is supplied.
- The React component is hard-coded to be rendered inside a `sprk-o-CenteredColumn`. We should make this optional instead of always having it inside one. It would be a breaking change to remove the class name. It should be optional though and the user could add it to `additionalClasses` and the component would function the same.
- Remove the empty string default props
- Fix the eslint errors
- Change the image option for `addClasses` to `additionalClasses`
- Change the image option for `linkAddClasses` to `linkAdditionalClasses`
- Add in oneOf for `splitAt` prop to define expected strings.
- Add in oneOf for `itemSpacing` prop to define the expected strings.
- Update disclaimer text to be My Disclaimer and not "My Disclaimer title"
- Add changed prop names to upgrade guides

bobwise:
## What does this PR do?
- Award will no longer render an empty Toggle
- Award will no longer render with `sprk-o-CenteredColumn` by default. This class can be added back with `additionalClasses` if it is desired.
- Property `addClasses` is deprecated. Use `imageAdditionalClasses` instead.
- Property `linkAddClasses` is deprecated. Use `linkAdditionalClasses` instead.
- Remove the empty string default props
- Fix the eslint errors
- Add in oneOf for splitAt prop to define expected strings.
- Add in oneOf for itemSpacing prop to define the expected strings.

### Associated Issue
Fixes #3315 

### Code
 - [X] Build Component in React
 - [X] Unit Testing in React with `npm run test` in `react/` (100% coverage, 100% passing)

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)
  
### Accessibility Testing
  - [x] Axe browser extension
  - [x] Jaws Inspect

### DIFF
--- a/react/src/components/awards/SprkAward.js
+++ b/react/src/components/awards/SprkAward.js
@@ -78,15 +78,17 @@ const SprkAward = (props) => {
         })}
       </SprkStack>
 
-      <SprkToggle
-        title={disclaimerTitle}
-        analyticsString={disclaimerAnalytics}
-        additionalClasses="sprk-o-Stack__item"
-      >
-        <p className="sprk-b-TypeBodyFour sprk-u-pts sprk-u-pbs">
-          {disclaimerText}
-        </p>
-      </SprkToggle>
+      {(disclaimerTitle || disclaimerText) && (
+        <SprkToggle
+          title={disclaimerTitle}
+          analyticsString={disclaimerAnalytics}
+          additionalClasses="sprk-o-Stack__item"
+        >
+          <p className="sprk-b-TypeBodyFour sprk-u-pts sprk-u-pbs">
+            {disclaimerText}
+          </p>
+        </SprkToggle>
+      )}
     </SprkStack>
   );
 };
--- a/react/src/components/awards/SprkAward.js
+++ b/react/src/components/awards/SprkAward.js
@@ -103,7 +103,11 @@ SprkAward.defaultProps = {
 };
 
 SprkAward.propTypes = {
-  /** Determines the spacing between the items. */
+  /**
+   * Determines the spacing between the items.
+   * `misc-a`, `misc-b`, etc are deprecated. Use `miscA`, `miscB`, etc
+   * instead.
+   */
   itemSpacing: PropTypes.oneOf([
     'tiny',
     'small',
--- a/react/src/components/awards/SprkAward.js
+++ b/react/src/components/awards/SprkAward.js
@@ -94,12 +94,6 @@ const SprkAward = (props) => {
 SprkAward.defaultProps = {
   splitAt: 'small',
   itemSpacing: 'medium',
-  additionalClasses: '',
-  idString: '',
-  heading: '',
-  disclaimerText: '',
-  disclaimerAnalytics: '',
-  disclaimerTitle: '',
 };
 
 SprkAward.propTypes = {
--- a/react/src/components/awards/SprkAward.js
+++ b/react/src/components/awards/SprkAward.js
@@ -82,7 +82,7 @@ const SprkAward = (props) => {
         })}
       </SprkStack>
 
-      {(disclaimerTitle || disclaimerText) && (
+      {disclaimerTitle && disclaimerText && (
         <SprkToggle
           title={disclaimerTitle}
           analyticsString={disclaimerAnalytics}
--- a/react/src/components/awards/SprkAward.js
+++ b/react/src/components/awards/SprkAward.js
@@ -122,7 +122,6 @@ SprkAward.propTypes = {
     'miscB',
     'miscC',
     'miscD',
-    '',
   ]),
   /**
    * Determines when the flex-direction should
@@ -135,7 +134,6 @@ SprkAward.propTypes = {
     'medium',
     'large',
     'huge',
-    '',
   ]),
   /**
    * A space-separated string of classes to add
--- a/react/src/components/awards/SprkAward.js
+++ b/react/src/components/awards/SprkAward.js
@@ -48,6 +48,8 @@ const SprkAward = (props) => {
             href,
             linkAddClasses,
             addClasses,
+            imageAdditionalClasses,
+            linkAdditionalClasses,
             alt,
             src,
             element,
@@ -58,7 +60,7 @@ const SprkAward = (props) => {
             <TagName
               className={classnames(
                 'sprk-o-Stack__item sprk-o-Stack__item--flex@s sprk-o-Stack',
-                linkAddClasses,
+                linkAdditionalClasses || linkAddClasses,
               )}
               href={TagName === 'a' ? href || '#nogo' : undefined}
               data-analytics={analyticsString}
@@ -68,7 +70,7 @@ const SprkAward = (props) => {
               <img
                 className={classnames(
                   'sprk-o-Stack__item sprk-o-Stack__item--center-column',
-                  addClasses,
+                  imageAdditionalClasses || addClasses,
                 )}
                 alt={alt}
                 src={src}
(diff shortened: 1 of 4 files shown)


react/src/components/awards/SprkAward.js
```diff
@@ -78,38 +82,57 @@ const SprkAward = (props) => {
...
-      <SprkToggle
-        title={disclaimerTitle}
-        analyticsString={disclaimerAnalytics}
-        additionalClasses="sprk-o-Stack__item"
-      >
-        <p className="sprk-b-TypeBodyFour sprk-u-pts sprk-u-pbs">
-          {disclaimerText}
-        </p>
-      </SprkToggle>
+      {(disclaimerTitle || disclaimerText) && (
```
RVMendoza, line 85: > Should update React component to not render the toggle if no toggle info is supplied.

If I had disclaimerTitle but no disclaimer text, then wouldn't an empty Toggle render?

I would expect this to need both Title and Text by using &&

react/src/components/awards/SprkAward.js
```diff
@@ -78,38 +82,57 @@ const SprkAward = (props) => {
...
+    'tiny',
+    'small',
+    'medium',
+    'large',
+    'huge',
+    'misc-a',
+    'misc-b',
+    'misc-c',
+    'misc-d',
+    'miscA',
```
afebbraro, line 117: I'm wondering if we deprecated one of the sets of values here. I don't think we want both misc-b and miscB for example.

afebbraro (changes requested):

react/src/components/awards/SprkAward.js
bobwise: Good question. I'll get some clarity on what the AC is going for.

react/src/components/awards/SprkAward.js
bobwise: i'll check

bobwise: It doesn't look like anything is deprecated currently. These are all still functional.

```
'sprk-o-Stack--misc-a': itemSpacing === 'miscA' || itemSpacing === 'misc-a',
'sprk-o-Stack--misc-b': itemSpacing === 'miscB' || itemSpacing === 'misc-b',
'sprk-o-Stack--misc-c': itemSpacing === 'miscC' || itemSpacing === 'misc-c',
'sprk-o-Stack--misc-d': itemSpacing === 'miscD' || itemSpacing === 'misc-d',
```

react/src/components/awards/SprkAward.js
bobwise: confirmed with @afebbraro - we're going to only show the Toggle if there is valid disclaimerText AND disclaimerTitle for both Angular and React.

react/src/components/awards/SprkAward.js
afebbraro: Thanks for checking! Looks like the hyphenated values should be deprecated since our naming format for props is camelCase. Lets get an issue in to deprecate those in stack and not reference them here.

bobwise: @afebbraro I created a new Issue to deprecate these prop values in `SprkStack`. I marked them as deprecated in `SprkAward` since they may currently be in use in production. Let me know what you think.

react/src/components/awards/SprkAward.js
```diff
@@ -78,38 +82,61 @@ const SprkAward = (props) => {
...
+    'miscB',
+    'miscC',
+    'miscD',
+    '',
```
guern1kn, line 125: Do we need these empty strings on `itemSpacing` and `splitAt`? They have defaults, so there shouldn't be a need to allow for an empty string.

react/src/components/awards/SprkAward.test.js
```diff
@@ -24,11 +24,10 @@ describe('SprkAward:', () => {
...
     const wrapper = mount(<SprkAward images={images} />);
-    expect(
-      wrapper.find(
-        'div.sprk-o-Stack.sprk-o-CenteredColumn.sprk-o-Stack--medium',
-      ).length,
-    ).toBe(1);
+    expect(wrapper.find('div').first().hasClass('sprk-o-Stack')).toBe(true);
+    expect(wrapper.find('div').first().hasClass('sprk-o-Stack--medium')).toBe(
+      true,
+    );
```
guern1kn, line 30: We should add something to this test to make it more explicit and make sure that no unnecessary classes are being added. A length check or find something to be false.

guern1kn (approved):

react/src/components/awards/SprkAward.js
afebbraro: Good call on marking them deprecated here.

afebbraro:
Good stuff! The only outstanding bit would be about do we want to add the `sprk-o-CenteredColumn` class to additionalClases in the the copy paste examples in SB? If we add it then the design won't change which is what we want. So I think we should that in the examples. To be even more helpful, we could add a line item to the SB docs for the award page that says that if they don't want it to be in the centered column they can remove that class from the additionalClasses.

bobwise:
> Good stuff! The only outstanding bit would be about do we want to add the `sprk-o-CenteredColumn` class to additionalClases in the the copy paste examples in SB? If we add it then the design won't change which is what we want. So I think we should that in the examples. To be even more helpful, we could add a line item to the SB docs for the award page that says that if they don't want it to be in the centered column they can remove that class from the additionalClasses.

Good catch! I'll add it

react/src/components/awards/SprkAward.js
```diff
@@ -137,10 +162,17 @@ SprkAward.propTypes = {
...
       alt: PropTypes.string.isRequired,
-      /** Additional classes for the image. */
+      /** Deprecated - Use `imageAdditionalClasses` instead.
+       * Additional classes for the image. */
       addClasses: PropTypes.string,
-      /** Additional classes for the link wrapping the image. */
+      /** Deprecated - Use `linkAdditionalClasses` instead.
+       * Additional classes for the link wrapping the image. */
       linkAddClasses: PropTypes.string,
+      /** Additional classes for the image. */
```
afebbraro, line 171: Hey @bobwise, Lets use a more descriptive comment description for our consumers. We use this one throughout the system for additional classes "A space-separated string of classes to add to the outermost container of the component." We change the part  of " to the outermost container" to something related to the specific use case. So, in this case you would change it to reflect that it is for the link wrapping the image.

react/src/components/awards/SprkAward.js
```diff
@@ -137,10 +162,17 @@ SprkAward.propTypes = {
...
+      imageAdditionalClasses: PropTypes.string,
+      /** Additional classes for the link wrapping the image. */
```
afebbraro, line 173: Same here for the comment description.

react/src/components/awards/SprkAward.js
bobwise: @afebbraro I'll go ahead and update these, but FYI they don't actually show up in storybook. Nested props like this are only displayed in this (inaccessible) tooltip without the doc comments.

image

afebbraro: @bobwise, They only don't show up right now. We want those comments to be consistent so that when we update Storybook (and the feature is added) we then have everything ready to go. Otherwise, we would have to go back and update all these comments.

afebbraro (approved):

PR #3552, opened 2020-10-06
React Button API Upgrades - separate Sprkpinner - Add isDisabled 

### LINKED ISSUE #3317: Button Upgrades - Part 2 (React)
**Notes:**
Angular Changes (this will be done in sparkdesignsystem/spark-design-system#3316 

**React Changes**
** Use deprecation strategy where we accept both props until the next major release**
- Change `disabled` to `isDisabled`
- Fix eslint errors

## Update
- Make this more like the angular #3316 
  - Compositional, take spinner out of SprkButton
- Leave existing `loading` prop on button, but add deprecation notice on dev docs.
- "engineering required"
- Hiding or showing the spinner is now the responsibility of the implementor
- Removing `loading` from sprk-button on next release after this one
- All existing buttons will continue to work until deletion of sprk-button's `loading` input happens
- Add new accessibility recommendations https://github.com/sparkdesignsystem/spark-design-system/issues/3569#issuecomment-713131375

RVMendoza:
## What does this PR do?
- [x] Change disabled to isDisabled
- [x] Fix eslint errors
- [x] Use deprecation strategy
- [x] Added relevant TODOs for removal
  - [x] Accessibility considerations
  - [x] disabled
- [x] Tests
  - [x] isDisabled
  - [x] Aria label tests 
- [ ] add to release docs

Additional Context:
https://github.com/sparkdesignsystem/spark-design-system/issues/3569
### Associated Issue
Fixes #3317
Followup removal: https://github.com/sparkdesignsystem/spark-design-system/issues/3557
### Documentation
 - [x] Update Spark Docs

### Code

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)

### DIFF
--- a/html/components/_buttons.scss
+++ b/html/components/_buttons.scss
@@ -5,7 +5,8 @@
 ////
 /// @group button
 ////
-.sprk-c-Button {
+.sprk-c-Button,
+.sprk-c-Button.sprk-c-Button--loading[disabled] {
   -webkit-appearance: none;
   background-color: $sprk-btn-background-color;
   border: $sprk-btn-border;
@@ -75,7 +76,8 @@
 /* stylelint-enable selector-class-pattern */
 
 /// Styles the button as the secondary variant.
-.sprk-c-Button--secondary {
+.sprk-c-Button--secondary,
+.sprk-c-Button--secondary.sprk-c-Button--loading[disabled] {
   background-color: $sprk-btn-secondary-background-color;
   border: $sprk-btn-secondary-border;
   color: $sprk-btn-secondary-text-color;
@@ -96,7 +98,8 @@
 }
 
 /// Styles the button as the tertiary variant.
-.sprk-c-Button--tertiary {
+.sprk-c-Button--tertiary,
+.sprk-c-Button--tertiary.sprk-c-Button--loading[disabled] {
   background-color: $sprk-btn-tertiary-background-color;
   box-shadow: none;
   border: $sprk-btn-tertiary-border;
@@ -118,7 +121,8 @@
 }
 
 /// Styles the button as the quaternary variant.
-.sprk-c-Button--quaternary {
+.sprk-c-Button--quaternary,
+.sprk-c-Button--quaternary.sprk-c-Button--loading[disabled] {
   background-color: $sprk-btn-quaternary-background-color;
   box-shadow: $sprk-btn-quaternary-box-shadow;
   border: $sprk-btn-quaternary-border;
--- a/react/src/components/buttons/SprkButton.js
+++ b/react/src/components/buttons/SprkButton.js
@@ -12,8 +12,8 @@ const SprkButton = ({
   isDisabled = disabled,
   element,
   idString,
-  isLoading,
   loading,
+  isLoading = loading,
   variant,
   href,
   spinningAriaLabel,
@@ -30,7 +30,7 @@ const SprkButton = ({
 
   // TODO: Remove anything related to spinner from button on next release #3557
   let spinnerVariant;
-  if (loading) {
+  if (isLoading) {
     if (variant === 'secondary') {
       spinnerVariant = 'primary';
     }
@@ -60,12 +60,12 @@ const SprkButton = ({
       href={TagName !== 'button' ? href : undefined}
       {...rest}
       // TODO: Remove loading on next release #3557
-      {...(loading && { 'aria-label': spinningAriaLabel })}
+      {...(isLoading && { 'aria-label': spinningAriaLabel })}
       {...(isLoading && { 'aria-live': 'polite' })}
     >
       {/* TODO: Remove anything related to spinner from button
       on next release #3557 */}
-      {(loading && <SprkSpinner variant={spinnerVariant} />) || children}
+      {(isLoading && <SprkSpinner variant={spinnerVariant} />) || children}
     </TagName>
   );
 };
--- a/react/src/components/buttons/SprkButton.js
+++ b/react/src/components/buttons/SprkButton.js
@@ -3,10 +3,12 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import SprkSpinner from '../spinners/SprkSpinner';
 
+// TODO: Remove loading, spinningAriaLabel and disabled on next release #3557
 const SprkButton = ({
   additionalClasses,
   analyticsString,
   children,
+  isDisabled,
   disabled,
   element,
   idString,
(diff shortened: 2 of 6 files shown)


react/src/components/buttons/SprkButton.test.js
```diff
@@ -159,6 +237,18 @@ describe('SprkButton:', () => {
...
+      ' and disabled is true',
+    () => {
+      const wrapper = mount(<SprkButton isDisabled="true" />);
+      const button = wrapper.find('button');
+      expect(button.length).toBe(1);
+      expect(button.getDOMNode().hasAttribute('disabled')).toBe(true);
+    },
+  );
+
+  // TODO: Deprecate
```
bobwise, line 251: Looks great!

Can we add a couple new tests to verify that the new Props are preferred over the deprecated Props?
```
<SprkButton isSpinning={true} loading={false} />

<SprkButton isDisabled={true} disabled={false} />
```

bobwise (changes requested):

react/src/components/buttons/SprkButton.test.js
RVMendoza: After our conversation a few minutes ago, the way booleans work, it's either there or it's not. If it's set to false, it's evaluated as not there at all.
There's no way to prefer the new prop over the old one because of the mentioned situation.

We're skipping writing these tests for now.

react/src/components/buttons/SprkButton.stories.js
```diff
@@ -115,13 +119,18 @@ export const disabledQuaternary = () => (
...
+    idString="button-9"
+    analyticsString="button-9-analytics"
+  >
     Button
   </SprkButton>
 );
 
 export const spinningSecondary = () => (
   <SprkButton
+    isSpinning
```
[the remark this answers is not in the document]
RVMendoza: good catch! done

react/src/components/buttons/SprkButton.js
```diff
@@ -73,11 +79,18 @@ SprkButton.propTypes = {
   analyticsString: PropTypes.string,
   /** Content to render inside of the SprkButton */
   children: PropTypes.node,
+  // TODO: Deprecate on next release
```
afebbraro, line 82: Hey @RVMendoza can you sync with Steph and get an issue in for the deprecate work and put the issue number here in the comments?

afebbraro (dismissed):

react/src/components/buttons/SprkButton.js
RVMendoza: @afebbraro this is finished and connected to issue that removes on next release https://github.com/sparkdesignsystem/spark-design-system/issues/3557

afebbraro:
Removing the PR review label since this is in progress based on the discussion from earlier about making the SprkSpinner something the consumer would add in themselves.

react/src/components/buttons/SprkButton.stories.js
```diff
@@ -32,7 +35,7 @@ alternative text must be provided.
 submit the form, the attribute \`type=”button”\` must be used.
 - If a Button submits a form, do not use \`<input type="submit">\`.
 Instead use \`<button type=”submit”>\`
-- If a Button is Disabled, you must add the \`disabled\` attribute
+- If a Button is Disabled, you must add the \`isDisabled\` attribute
```
bobwise, line 38: I think this should stay as `disabled`.

Also doesn't this only apply to the HTML framework? Angular and React should handle this automatically?

bobwise (changes requested):

react/src/components/buttons/SprkButton.stories.js
RVMendoza: this story is for React. We use `isDisabled` when referring to something that's disabled so this is part of making it consistent.

so it would look like this `<SprkButton isDisabled>Text</SprkButton>`

Could you clarify what you mean?

bobwise: If I understand correctly, this bullet is guidance specifically for the HTML framework. For HTML Buttons, devs may think that disabling a button just involves adding the `sprk-is-Disabled` class, since doing so makes the button look disabled. However, devs must also add the HTML `disabled` attribute to achieve the desired behavior.

If I'm understanding that bullet correctly, that would mean this guidance is just noise for Angular and React devs and this whole bullet should be removed from those two frameworks.

RVMendoza: I see okay, this guidance is noise for angular and react and is not relevant. Probably copied from HTML. Will look through this dev doc and get rid of stuff that's not important.

html/components/_buttons.scss
```diff
@@ -5,7 +5,8 @@
 ////
 /// @group button
 ////
-.sprk-c-Button {
+.sprk-c-Button,
+.sprk-c-Button.sprk-c-Button--loading[disabled] {
```
RVMendoza, line 9: this is done because we are incorrectly styled based on a disabled attribute. So I have to be more specific instead of undoing everything. 

We've all agreed to not style based on attribute or elements but haven't gotten around to doing it here. I'm making an issue to do for later because this would be a breaking change otherwise.

react/src/components/buttons/SprkButton.js
```diff
@@ -43,17 +48,25 @@ const SprkButton = ({
...
+        { 'sprk-c-Button--loading': isLoading },
         additionalClasses,
       )}
       role={TagName !== 'button' ? 'button' : undefined}
       data-id={idString}
       data-analytics={analyticsString}
-      disabled={TagName !== 'a' ? disabled : undefined}
+      // TODO: Remove disabled on next release #3557
+      disabled={
+        TagName !== 'a' ? isDisabled || disabled || isLoading : undefined
```
RCopeland, line 60:
```suggestion
        TagName !== 'a' ? isDisabled || isLoading : undefined
```

RCopeland (approved):
suggested a cleaner deprecation for disabled, otherwise lgtm

react/src/components/buttons/SprkButton.test.js
RCopeland: the suggestion i made in there should set it up so that the new one is preferred

RVMendoza:
@RCopeland woah I didn't even realize you could do that!!! Added those commits in

guern1kn (approved):

Issue #3461, opened 2020-09-23
ESLint fix JS files for HTML


afebbraro:
Fix any eslinting errors in the JS files in the HTML directory

devcshort:
Can I do this once hacktoberfest starts?

bobwise:
Hello @devcshort !

You can absolutely work on this whenever you'd like :)

devcshort:
Sounds great! Can you assign this one to me?

devcshort:
I haven't heard anything back, so I'm going to drop out of this one. I also don't have much time anymore due to a new job. Thank you!

motionsuggests:
I will take this one.

motionsuggests:
So I'm not sure what to do about these `max-len` errors. Breaking them feels weird. Should I just put in `// eslint-disable-next-line max-len` so it stops complaining?

Example
```
<div class="sprk-o-Stack sprk-o-Stack--split@xs">
      <div class="sprk-o-Stack__item sprk-o-Stack__item--fourth@xs sprk-u-AbsoluteCenter"> // complaining here
        <p class="sprk-b-TypeBodyOne">
          fourth
        </p>
      </div>
      <div class="sprk-o-Stack__item sprk-o-Stack__item--half@xs sprk-u-AbsoluteCenter"> // complaining here
        <p class="sprk-b-TypeBodyOne">
          half
        </p>
      </div>
      <div class="sprk-o-Stack__item sprk-o-Stack__item--fourth@xs sprk-u-AbsoluteCenter"> // complaining here
        <p class="sprk-b-TypeBodyOne">
          fourth
        </p>
      </div>
    </div>
```

PR #3610, opened 2020-10-29
Fix most of the ESLint errors for JS files in HTML

motionsuggests:
## What does this PR do?
This fixes most of the eslint errors for `/html`.  The remaining errors probably require more understanding of how the code is used. 

## Couple caveats
- I disabled `max-len` for max-length lines because the lines that were breaking the rule were class attributes that when displayed in Storybook would look weird to break.
- I disabled `global-require` which does not allow using something like `const { highlightBoard } = require('../components/highlight-board');` inside of a function because it breaks the tests and I'm not sure there is a simple solution for this. https://eslint.org/docs/rules/global-require

### Associated Issue
https://github.com/sparkdesignsystem/spark-design-system/issues/3461

### Code
 - [x] Unit Testing in HTML with `npm run test` in `html/` (100% coverage, 100% passing

## Remaining eslint errors/warnings
```
/Users/jtaylor17/code/github/spark-design-system/html/utilities/generateAriaControls.js
  16:5  warning  Unexpected console statement  no-console
  24:5  warning  Unexpected console statement  no-console

/Users/jtaylor17/code/github/spark-design-system/html/utilities/generateIdForInput.js
   9:5  warning  Unexpected console statement                                 no-console
  17:5  warning  Unexpected console statement                                 no-console
  30:3  error    Assignment to property of function parameter 'labelElement'  no-param-reassign

/Users/jtaylor17/code/github/spark-design-system/html/utilities/isElementVisible.js
  11:3  error  Arrow function expected no return value  consistent-return

/Users/jtaylor17/code/github/spark-design-system/html/utilities/polyfills/CustomEvent.js
  1:2  warning  Unexpected unnamed function                        func-names
  1:2  error    Expected to return a value at the end of function  consistent-return
  5:5  error    Assignment to function parameter 'params'          no-param-reassign

/Users/jtaylor17/code/github/spark-design-system/html/utilities/polyfills/ObjectAssign.js
  14:28  error  Use the rest parameters instead of 'arguments'                                                                                                                         prefer-rest-params
  17:11  error  for..in loops iterate over the entire prototype chain, which is virtually never what you want. Use Object.{keys,values,entries}, and iterate over the resulting array  no-restricted-syntax

/Users/jtaylor17/code/github/spark-design-system/html/utilities/polyfills/ObjectPrepend.js
   2:2   warning  Unexpected unnamed function                                                func-names
   3:15  warning  Unexpected unnamed function                                                func-names
   4:14  error    Do not access Object.prototype method 'hasOwnProperty' from target object  no-prototype-builtins
  12:51  error    Use the rest parameters instead of 'arguments'                             prefer-rest-params
  15:24  warning  Unexpected unnamed function                                                func-names

/Users/jtaylor17/code/github/spark-design-system/html/utilities/scrollYDirection.js
  10:3  error  Arrow function expected no return value  consistent-return

✖ 17 problems (9 errors, 8 warnings)
```


afebbraro:
Thanks for the PR @motionsuggests , we'll review this soon.

.eslintrc
```diff
@@ -29,5 +29,19 @@
...
     }]
-  }
+  },
+  "overrides": [
+    {
+      "files": ["html/**/*.js"],
+      "rules": {
+        "max-len": "off"
+      }
+    },
```
guern1kn, line 39: Thanks for the PR, @motionsuggests ! We don't want to disable max-len for all the files. We take it on more of an exception basis than as an overall rule. Would you mind removing this and updating with staging? Everything else looks good!

guern1kn (changes requested):

.eslintrc
motionsuggests: Sorry for not getting to this sooner. Should I just go through and disable max-len inline where necessary? Like this
```js
export default {
  title: 'Components/Input/Checkbox',
  decorators: [(story) => `<div class="sprk-o-Box">${story()}</div>`],
  /* eslint-disable max-len */
  parameters: {
    info: `
${markdownDocumentationLinkBuilder('input')}
- Elements containing role="checkbox" must also include the
 aria-checked attribute to expose the checkbox's state to assistive technology.
- Since a checkbox is an interactive control, it must be focusable
and keyboard accessible. If the role is applied to a non-focusable element,
use the tabindex attribute to change this.
- Change the value of the aria-checked attribute
dynamically when the checkbox is activated.

###### Ensure all related checkboxes are grouped together using one of the following methods:

- Fieldset and legend tags
- ARIA groups role="group" and aria-label or aria-labelledby
- aria-labelledby that points to the same
 element for every checkbox in the group
    `,
    docs: { iframeHeight: 180 },
  },
};
/* eslint-enable max-len */
```

motionsuggests: @guern1kn This is ready for review. I went ahead and updated everywhere to use 
```
/* eslint-disable max-len */
/* eslint-enable max-len */
```

RVMendoza: Hi @motionsuggests thank for all you work on this so far.

So for places like 
```
###### Ensure all related checkboxes are grouped together using one of the following methods:
```

would turn into:
```
###### Ensure all related checkboxes are grouped together
using one of the following methods:
```

We typically break the line into the next to avoid excessive line length. The Gatsby parses it fine as long as it's immediately on the next line.

--------

**I'm also seeing a lot of places in this PR that is disabling max-len. We're looking for a change in code and disabling max-len as a last resort.**

For example: 

dropdown.stories line 97
```
ESLint doesn't like this for max length
        <svg
          class="sprk-c-Icon sprk-c-Icon--filled-current-color sprk-c-Icon--stroke-current-color sprk-u-mls"
          viewBox="0 0 100 100">
          <use xlink:href="#chevron-down" />
        </svg
```

Would turn into:
```
        <svg
          class="
            sprk-c-Icon 
            sprk-c-Icon--filled-current-color 
            sprk-c-Icon--stroke-current-color
            sprk-u-mls
          "
          viewBox="0 0 100 100">
          <use xlink:href="#chevron-down" />
        </svg>
```

##  💡 **Please go through where you disabled and see if you can correct this way. If it can't be fixed, disable max-len is fine, but it should be a last resort.**

motionsuggests: @RVMendoza Thanks! This makes sense. I was concerned about breaking lines for `class=` because it is used in the code examples but this solves those.

motionsuggests:
@guern1kn @RVMendoza  I think I got it this time 😃 .

lego movie gif - I wasn't listening

guern1kn (approved):

guern1kn:
@motionsuggests Thank you for fixing these!

PR #3727, opened 2020-12-23
Resolve missing SCSS module

lovettbp:
This seems to have (accidentally?) changed from the v13 docs.

Left as is, page build will fail with the following error:

```
ERROR in ./src/index.js 7:0-47
Module not found: Error: Can't resolve '@sparkdesignsystem/spark-styles/spark' in 'C:\Users\blovett\GitHub\Sorry\spark-html\src'

webpack 5.11.0 compiled with 1 error in 3956 ms

```
Reverting this import line to its v13 state allows the page to build without issue.

Windows 10 20H2 (19041), PowerShell 7.2.

### DIFF
--- a/src/pages/installing-spark/html-installation.mdx
+++ b/src/pages/installing-spark/html-installation.mdx
@@ -44,7 +44,7 @@ import sparkPrerender from "@sparkdesignsystem/spark/es5/sparkPrerender";
 import "@sparkdesignsystem/spark/es5/sparkPolyfills";
 
 // import the Spark CSS
-import "@sparkdesignsystem/spark-styles/spark";
+import "@sparkdesignsystem/spark-styles/_spark.scss";
 
 // initialize Spark
 sparkPrerender();


afebbraro:
Hey @lovettbp! Thanks for reaching out. For Sass, you do not need to include the leading underscore or the file extensions. From the Sass docs:

"you don’t have to explicitly write out the extension of the file you want to import; @import "variables" will automatically load variables.scss, variables.sass, or variables.css."

"These are called partials, and they tell Sass tools not to try to compile those files on their own. You can leave off the _ when importing a partial."

https://sass-lang.com/documentation/at-rules/import

I'm curious why your build setup is throwing an error without those two things. Do you have a link to your setup so we can check it out or could you paste some things here and we can try to help.

afebbraro (changes requested):
Left comment for discussion.

afebbraro:
Closing for now. We talked in chat over this. Will re-open if necessary.

Issue #3585, opened 2020-07-24
[Enhance] Fix axe Error - React, Inputs


ssabo1:
**User Story:**
SprkSelectionInput is adding aria-describedBy to inputs even if they don't exist. This is an axe error in React. Possibly Angular/HTML as well.

**Notes:**
SME: Bob

**AC:**
- will not see Axe error in React
- will not see in HTML?
- will not see axe error in Angular??

bobwise:
Closed with https://github.com/sparkdesignsystem/spark-design-system/pull/3784

PR #3784, opened 2021-01-19
SprkTextInput aria-describedby

### LINKED ISSUE #3526: SprkInputElement should only have aria-describedby if the error container exists
## Describe The Bug
`SprkInputElement` is hard-coded to have aria-describedby even if the id provided does not match an existing error container. This causes an a11y violation.

## To Reproduce
1. Go to https://react.sparkdesignsystem.com/iframe.html?id=components-input-text--text-input
2. Open Chrome Dev Tools > Audits
3. Click "Generate Report"

Lighthouse report:
image

Source code:
image

## Expected Behavior
- SprkInputElement should only render with aria-describedby if the id is valid.
- New unit tests that test for this functionality

bobwise:
~~Modifies SprkTextInput to only include `aria-describedby` on the `<input>` if it has an error container.~~

New scope: Refactor `SprkTextInput` and `SprkInputElement` to support these use cases:
1. User can set their own value(s) for `aria-describedby` on the input.
2. If the input is invalid, the component will merge the error container ID with any values provided by the user.
3. `aria-describedby` will only include the error container ID if the value is invalid.
4. `aria-describedby` will only be calculated if `errorContainerId` or `ariaDescribedBy` have been updated.

### Associated Issue
Fixes #3526 (also https://github.com/sparkdesignsystem/spark-design-system/issues/3585)

### Code

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)
  
### Accessibility Testing
  - [x] Axe browser extension
  - [x] Jaws Inspect

 - [x] Accessibility Team


afebbraro (changes requested):
aria-described by could be used for more use cases besides the field being invalid. If we only allow it if there’s an error container then we could be restricting use. What do you think?

guern1kn (approved):
I think that @afebbraro has a good point, but I think it's fine the way it currently is. We are currently setting the `errorContainerId` internally and passing it to the `<SprkInputElement>`. Both the `<SprkTextInput>` and `<SprkInputElement>` are in the process of being replaced. I don't know that it makes sense to spend time re-writing these components to allow for additional functionality at this point. Amber's point would be something to keep in mind for the React Input API upgrades though.

bobwise:
Yes @afebbraro is correct. `aria-describedby` can actually have multiple values passed into it at the same time, which Spark should support. Spark should be putting the helper text container id in the `aria-describedby` as well as allowing users to add their own. This is one of the items that came up in our accessibility audit (HTML framework). https://github.com/sparkdesignsystem/Backlog/issues/1332

react/src/base/inputs/SprkTextInput/SprkTextInput.js
```diff
@@ -67,7 +67,7 @@ class SprkTextInput extends Component {
               forwardedRef={forwardedRef}
               type={type}
               formatter={formatter}
-              errorContainerId={errorContainerId}
+              errorContainerId={valid ? undefined : errorContainerId}
```
RVMendoza, line 70: Nice! This is a good way to make sure screen readers aren't confused if the error doesn't exist.

RVMendoza (approved):

afebbraro:
@bobwise  If we want to fix the bug and not introduce new a11y errors, it would be best to update this component to accept an ariaDescribedBy prop that if it has a value would pass down to the input element component and additionally, if the input is also invalid it would join errorContainerId and the ariaDescribedBy into a string that it would pass down. That away, users can supply their own ariaDescribedBy and we would also linking the error container correctly.

guern1kn (approved):

RVMendoza (approved):

react/src/base/inputs/components/SprkInputElement/SprkInputElement.js
```diff
@@ -51,6 +52,18 @@ class SprkInputElement extends Component {
       }
     };
 
+    const getAriaDescribedBy = () => {
```
RVMendoza, line 55: nice! this is definitely something we should be incorporating to our other components that use aria-describedby

react/src/base/inputs/components/SprkInputElement/SprkInputElement.js
```diff
@@ -73,9 +86,9 @@ class SprkInputElement extends Component {
         data-id={idString}
         data-analytics={analyticsString}
         aria-invalid={!valid}
-        aria-describedby={errorContainerId}
+        aria-describedby={getAriaDescribedBy()}
```
afebbraro, line 89: Thanks for adding this. I did notice a potential issue. This seems to be running every time the component renders for me, so even though the errorcontainer /id hasnt changed this gets triggered. Every time I type a character in the input, it runs. Which seems to drag down performance. Maybe I am doing something weird? But thats what I am seeing.  I think something to try here would be to bind this to a method that runs on one of the lifecycle hooks, componentDidMount() ?

afebbraro (approved):

PR #3802, opened 2021-01-25
remove sprk-u-Width-100 class from SprkInput component

guern1kn:
## What does this PR do?
Remove width utli from SprkInput


afebbraro (changes requested):
Thanks for doing this! I'm thinking It would make it easier on us if we update the CSS right now instead of having to add all these additional classes and then remove them again. We can get an issue and have an engineer that is free do that so you can continue to focus on inputs.

PR #3912, opened 2021-03-09
Add React Card Upgrades

afebbraro:
## What does this PR do?
This is a breaking change. In order for consumers to go to this new version, the old cards will have to be redone following the new story layout. This upgrade entails making the cards compositional by allowing content to project through. The available props now are isStandout to add the shadow standout styles and idString and additionalClasses. The React cards now match Angular cards. We can easily build new card patterns now without having to do a package release.  

### Associated Issue
Fixes React Card Issues

### Documentation
 - [x] Update Spark Docs

### Code

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)
  
### Accessibility Testing
  - [x] Axe browser extension
  - [x] Keyboard navigated to links
  - [ ] Jaws Inspect
  - [ ] VoiceOver (iOS) or Talkback (Android)


react/src/components/card/components/SprkCardHighlightedHeader/SprkCardHighlightedHeader.js
```diff
@@ -1,64 +0,0 @@
-import React from 'react';
```
afebbraro, line 1: Checked in spark-exports-react and we weren't exporting these internal components so there was no removal required from those files for these removed components.

react/src/components/card/SprkCard.js
```diff
@@ -1,55 +1,19 @@
...
+  const { additionalClasses, children, idString, isStandout, ...other } = props;
 
   return (
     <div
       className={cx('sprk-c-Card sprk-o-Stack', additionalClasses, {
         'sprk-c-Card--standout': isStandout,
-        'sprk-u-TextAlign--center': teaserConfig.media.mediaVariant === 'icon',
       })}
       data-id={idString}
+      {...other}
```
afebbraro, line 14: We weren't spreading the other props before so I added that in as a new feature.

react/src/components/card/SprkCard.stories.js
```diff
@@ -11,18 +15,28 @@ export default {
   ],
   component: SprkCard,
   parameters: {
-    subcomponents: { SprkCardTeaser, SprkCardHighlightedHeader },
+    subcomponents: {
+      SprkHeading,
+      SprkText,
+      SprkStack,
+      SprkLink,
+      SprkStackItem,
```
afebbraro, line 23:
```suggestion
      SprkStackItem,
      SprkIcon,
```

react/src/components/card/SprkCard.js
```diff
@@ -61,165 +25,23 @@ SprkCard.propTypes = {
...
-     * Text that will show as as
-     * the main headline of the Teaser Card.
-     */
-    title: PropTypes.string,
-    /**
-     * Determines if the Teaser Card's
-     * title is at the top or below the main media.
-     */
-    titleFirst: PropTypes.bool,
-  }),
```
RVMendoza, line 208: tenor-197725802

react/src/components/card/SprkCard.js
```diff
@@ -1,55 +1,19 @@
...
-      >
-        {children}
-      </div>
-    );
-  };
+  const { additionalClasses, children, idString, isStandout, ...other } = props;
 
   return (
     <div
       className={cx('sprk-c-Card sprk-o-Stack', additionalClasses, {
```
RVMendoza, line 10: Leaving sprk-o-Stack here would imply that someone would only ever put stack items in a card.

I'm okay with if we feel it's a reasonable assumption to make--I'm just thinking about a situation where someone might want to put something in a Card that's not a Stack Item

afebbraro: @RVMendoza, I had a similar line of thought as well. But, ultimately, the Card is a layout pattern that needs the flexbox layout system. If this is omitted, the card's layout wouldn't render correctly. So, it needs to be there and it's more of a requirement that all the items are flexbox items. Because of the display: flex setting, the items in the card don't technically need to be official StackItems. They will inherit the base flex properties regardless. So our bases are covered here and we'll leave this class on there.

react/src/components/card/SprkCard.stories.js
```diff
@@ -11,18 +15,29 @@ export default {
...
 
 export const defaultStory = () => (
-  <SprkCard
-    idString="card-1"
-    additionalContentClasses="sprk-o-Stack sprk-o-Stack--large"
-  >
-    Base Card Content
+  <SprkCard idString="default">
+    <SprkStack
+      additionalClasses="sprk-o-Stack__item sprk-c-Card__content"
```
RVMendoza, line 34: Related to [this comment]( https://github.com/sparkdesignsystem/spark-design-system/pull/3912/files#r591824056)

SprkCard contains the parent container that relates to Stack__item. 

My instinct here is to make this less opinionated and to instead move the SprkStack so then the user can easily understand where the `sprk-o-Stack__item` is related to.

```
<SprkCard>
     <SprkStack
          <SprkStack additionalClasses="sprk-o-Stack__item sprk-c-Card__content"
               <SprkStackItem
               <SprkStackItem
</SprkCard>
```

afebbraro: responded up above, resolving this one

react/src/components/card/SprkCard.js
RVMendoza: So no one can ever use Card like this even if they don't intend to use flex styles?
```
<SprkCard>
   <div class="sprk-c-Card__content">
      <h2>Hello</h2>
      <p>I love cats</p>
   </div>
</SprkCard>
```

This implementation would receive unnecessary Stack styles that the user might not immediately know about.
We'd be boxing them into a flex container even if they don't want/need it.

It feels like if we keep it this way, we'd need to communicate that Cards will be a flex container and there's no way to choose otherwise.

afebbraro: They can absolutely use that code that you pasted. It does feel that you may be picking at the code here with saying that we have to communicate that cards are flex containers and that we are restricting them otherwise. As with our other components, any of our styles are overridable via additionalClasses (display: flex cascading throughout a component would be comparable to font styles). In addition, our alerts, inputs, accordions, dictionary, footer, tooltip parts, and many other components all have display: flex on them. Almost all of our components are flex containers but we don't explicitly describe that CSS property for each component.

react/src/components/card/SprkCard.js
bobwise: 🔥

react/src/components/card/SprkCard.test.js
```diff
@@ -8,109 +7,39 @@ import SprkCard from './SprkCard';
 Enzyme.configure({ adapter: new Adapter() });
 
 describe('SprkCard:', () => {
```
bobwise, line 9: I love how much cleaner everything looks without the config objects in the code!

Do we want to add tests for `additionalClasses`?

bobwise (changes requested):
Great job on this! I think we can add another unit test but other than that lgtm.

react/src/components/card/SprkCard.test.js
afebbraro: Thanks @bobwise ! Great idea, I will add the test and update HTML to use secondary.

react/src/components/card/SprkCard.test.js
```diff
@@ -8,109 +7,39 @@ import SprkCard from './SprkCard';
...
-          description: 'Highlight Header Description',
-        }}
-      />,
-    );
-    const loadsHighlightedHeaderComponent =
-      wrapper.find('GetCardContent SprkCardHighlightedHeader').length === 1;
-    expect(loadsHighlightedHeaderComponent).toBe(true);
+  it('should not assign data-id when idString is not provided', () => {
+    const wrapper = mount(<SprkCard />);
+    expect(wrapper.find('[data-id="321"]').length).toBe(0);
```
guern1kn, line 43: Update to check to see if they `data-id` attribute exists. Not necessarily this specific value.

guern1kn (changes requested):
Just one comment. This PR makes me very happy. Ha.

react/src/components/card/SprkCard.test.js
afebbraro: Good catch, updated!

bobwise (approved):

guern1kn (approved):

PR #3919, opened 2021-03-15
Update Default Link Color

### LINKED ISSUE #3918: Update Default Link hover color will be Red
**User Story:**

**Notes:**

**AC:**
- Default Link underline will go from grey to purple
- on hover focus active, text and underline will be red

bobwise:
This PR updates various styles and hover/focus styles.

### Default Link
  * Default state is purple text, purple underline
  * Visited state is purple text, purple underline
  * Hover/Focus/Active state is red text, red underline

image

### Icon With Text Link
  * styles match Default Link but without an underline
  * Adds the link transition to the icon so it animates the same as the text.

image

### Informational Dropdown
  * text and chevron turn purple on hover
  * applies to both open and closed state

image

### Footer Toggle (ONLY footer)
  * text and icon will be bold and underlined on hover, like Footer Links.
  
image

### Masthead
  * Masthead Selector text, border, and chevron icon will turn purple on hover when the selector is open and when it is closed
  * Updated a couple spots in HTML and Angular stories where we were still using Dropdown instead of Masthead Selector.
  * Extended navigation dropdown items will turn red on hover

image

image

### Associated Issue
Fixes #3918 

### Code
 - [x] Update styles

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)
  
### Accessibility Testing
  - [x] Axe browser extension

 - [x] Design Team


styles/base/_links.scss
```diff
@@ -72,20 +72,27 @@
...
   color: $sprk-link-has-icon-color-icon;
   fill: $sprk-link-has-icon-fill;
   stroke: $sprk-link-has-icon-stroke;
+  transition: $sprk-link-transition;
 }
 
-.sprk-b-Link--has-icon:hover .sprk-c-Icon {
+.sprk-b-Link--has-icon:hover .sprk-c-Icon,
+.sprk-b-Link--has-icon:focus .sprk-c-Icon,
+.sprk-b-Link--has-icon:active .sprk-c-Icon {
```
bobwise, line 95: We were missing the rules for `:focus` and `:active`, which caused the icon color to not match the text color.

styles/base/_links.scss
```diff
@@ -72,20 +72,27 @@
...
+    border-bottom: $sprk-link-has-icon-visited-border-bottom;
+    color: $sprk-link-has-icon-visited-color-text;
+  }
 }
 
 .sprk-b-Link--has-icon .sprk-c-Icon {
   color: $sprk-link-has-icon-color-icon;
   fill: $sprk-link-has-icon-fill;
   stroke: $sprk-link-has-icon-stroke;
+  transition: $sprk-link-transition;
```
bobwise, line 90: Added this so the icon and text would animate together instead of changing color at different times.

guern1kn (approved):

styles/base/_links.scss
```diff
@@ -72,20 +72,27 @@
...
   &:hover,
   &:active,
-  &:focus,
-  &:visited {
+  &:focus {
     border-bottom: $sprk-link-has-icon-hover-border-bottom;
     color: $sprk-link-has-icon-hover-color-text;
   }
+
+  &:visited {
```
RVMendoza, line 80: probably a good thing you separated them for better customizability as well

RVMendoza (approved):
✨

afebbraro (changes requested):
Needs changes, discussed in chat. Thanks!

angular/projects/spark-angular/src/lib/components/sprk-masthead/sprk-masthead-selector/sprk-masthead-selector.component.ts
```diff
@@ -13,8 +13,8 @@ import { ISprkMastheadSelectorChoice } from '../sprk-masthead-selector/sprk-mast
 @Component({
   selector: 'sprk-masthead-selector',
   template: `
-    <div [ngClass]="{ 'sprk-c-MastheadMask': isOpen }">
-      <div class="sprk-o-Box">
+    <div [ngClass]="{ 'sprk-c-MastheadMask': isOpen && isFlush }">
+      <div [ngClass]="{ 'sprk-o-Box': isFlush }">
```
bobwise, line 17: `isFlush` was part of the React Masthead Selector but wasn't part of Angular yet.

angular/projects/spark-angular/src/lib/components/sprk-masthead/sprk-masthead-selector/sprk-masthead-selector.component.ts
```diff
@@ -51,7 +51,7 @@ import { ISprkMastheadSelectorChoice } from '../sprk-masthead-selector/sprk-mast
           <a
             sprkLink
             variant="plain"
-            class="sprk-o-Stack sprk-o-Stack--split@xxs sprk-o-Stack--center-column sprk-u-Width-100"
+            class="sprk-o-Stack sprk-o-Stack--split@xxs sprk-o-Stack--center-column sprk-u-Width-100 sprk-c-Masthead__selector-dropdown-header-link"
```
bobwise, line 54: This is a new class used to apply styles to these links.

afebbraro: The sprk-u-Width-100 is not needed here because Stack applies width 100%.

styles/tokens/footer.json
```diff
@@ -66,14 +66,34 @@
...
+          "file": "settings",
+          "themable": true
         },
-        "file": "settings",
-        "themable": true
+        "font-weight": {
+          "comment": "Toggle trigger hover font-weight on the Footer.",
+          "value": "500",
+          "attributes": {
+            "category": "content"
```
afebbraro, line 83: This is a number and not a string of content. Could use size here.

angular/projects/spark-angular/src/lib/components/sprk-masthead/sprk-masthead-selector/sprk-masthead-selector.component.ts
afebbraro: Nice. Thank you for adding this! I feel bad having to mention that we also need tests for this but we do haha. Good catch!

styles/tokens/footer.json
bobwise: `500` is a number but there are other values for `font-weight` that are not numbers. Does that mean they should all be `content` or should the numeric ones be switched to `size`? If I change them all to `size` I get compile errors.

afebbraro: It matters what we have the value as because when we build the distributed token files with these values. So, if we have the font weight value as "bold" that's a string and should be set as content. If we have the value as 500 like we do here, it should be size since it's a number.

bobwise: got it. Thanks!

afebbraro: awesome, good work!

afebbraro (approved):
Great work!

angular/projects/spark-angular/src/lib/components/sprk-masthead/sprk-masthead-selector/sprk-masthead-selector.component.ts
```diff
@@ -221,6 +221,9 @@ export class SprkMastheadSelectorComponent implements OnChanges {
    */
   @Input()
   triggerText: string;
+  /** Applies styles if the selector is flush with the sides of the viewport. */
+  @Input()
+  isFlush: boolean;
```
RVMendoza, line 226: I'm not gonna split hairs here: Renaming this to something that could be more descriptive as to what it does could make this easier to understand what it's doing in the long run.

- `isNarrowLayout`
- `hasNarrowStyles`

or something like that.

PR #3958, opened 2021-04-07
Adjust footer link font weight

bobwise:
Modifies the `font-weight` of Footer Links to be the same in the default and hover/focus states so that the text will not change size when it is interacted with.


afebbraro (approved):
looks good!

RVMendoza (approved):

PR #3973, opened 2021-04-23
Input Icon Alt Text

bobwise:
## What does this PR do?
- Adds `aria-hidden="true"` and `focusable="false"` to all input icons and error container icons in HTML and Angular.
- Autocomplete, Search, Percentage, Date Picker, and Select. (Monetary is not an SVG).

### Associated Issue
Fixes 2303668

### Code
 - [x] Build Component in HTML

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)
  
### Accessibility Testing
  - [x] Axe browser extension
  - [x] Jaws Inspect

 - [x] Accessibility Team

### DIFF
--- a/angular/projects/spark-angular/src/lib/components/inputs/DatePicker.stories.ts
+++ b/angular/projects/spark-angular/src/lib/components/inputs/DatePicker.stories.ts
@@ -79,6 +79,8 @@ export const defaultStory = () => ({
           iconName="calendar"
           additionalClasses="sprk-c-Icon--stroke-current-color"
           sprk-input-icon
+          ariaHidden="true"
+          focusable="false"
         ></sprk-icon>
         <input
           id="datepicker"
@@ -120,6 +122,8 @@ export const invalidDatePicker = () => ({
           iconName="calendar"
           additionalClasses="sprk-c-Icon--stroke-current-color"
           sprk-input-icon
+          ariaHidden="true"
+          focusable="false"
         ></sprk-icon>
         <input
           id="invalid-datepicker"
@@ -139,6 +143,8 @@ export const invalidDatePicker = () => ({
         <sprk-icon
           iconName="exclamation-filled"
           additionalClasses="sprk-b-ErrorIcon"
+          ariaHidden="true"
+          focusable="false"
         ></sprk-icon>
         <div class="sprk-b-ErrorText">There is an error on this field.</div>
       </span>
@@ -171,6 +177,8 @@ export const disabledDatePicker = () => ({
           iconName="calendar"
           additionalClasses="sprk-c-Icon--stroke-current-color"
           sprk-input-icon
+          ariaHidden="true"
+          focusable="false"
         ></sprk-icon>
         <input
           id="disabled-datepicker"
@@ -212,6 +220,8 @@ export const legacyStory = () => ({
         iconType="calendar"
         additionalClasses="sprk-c-Icon--filled-current-color sprk-c-Icon--stroke-current-color"
         sprk-input-icon
+        ariaHidden="true"
+        focusable="false"
       ></sprk-icon>
       <input
         id="legacy-datepicker"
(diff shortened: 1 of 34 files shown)


angular/projects/spark-angular/src/lib/components/inputs/DatePicker.stories.ts
```diff
@@ -79,6 +79,8 @@ export const defaultStory = () => ({
           iconName="calendar"
           additionalClasses="sprk-c-Icon--stroke-current-color"
           sprk-input-icon
+          ariaHidden="true"
+          focusable="false"
```
RVMendoza, line 83: I'm wondering if there's ever a situation where this would need to be different.

Either way, I'm fine with giving people the option and making it a default.

bobwise: I can't think of a scenario where we would want to make the svg focusable (better to wrap it in a button at that point). But I can definitely think of scenarios where they may want the icon to not be hidden and to have an aria-label instead.

Here are some examples I came up with while thinking about this story.

https://codepen.io/bobwise/full/zYNaPVZ

RVMendoza (approved):
This makes sense, icons on inputs are not meaningful elements, they're just decorations. Good work!

afebbraro (changes requested):
This is a nice addition. Thanks for adding it. I only had one small thing to mention and its the `focusable` name for the input. We use `is` or `has` in our naming convention so this one would be `isFocusable` to line up with that. Happy to chat more if you want.

bobwise:
> This is a nice addition. Thanks for adding it. I only had one small thing to mention and its the `focusable` name for the input. We use `is` or `has` in our naming convention so this one would be `isFocusable` to line up with that. Happy to chat more if you want.

Thanks @afebbraro! I agree that `isFocusable` is in line with our naming convention, except that setting this attribute to true does not make the SVG focusable.

`focusable` is a deprecated attribute in the SVG specification that is [only supported by IE 11](https://stackoverflow.com/questions/29393144/how-to-prevent-svg-elements-from-gaining-focus-with-tabs-in-ie11). Best practice is to set `focusable="false"` as a polyfill for IE 11. I called the input `focusable` so it would map directly to the SVG attribute, but I understand why that is confusing.

What about something like `ie11Focusable` or `svgFocusablePolyfill`?

bobwise:
After team discussion, we're going to add two changes:

1. `focusable` will be renamed to `focusableAttr`
2. The Input will default to `false` so we do not need to explicitly set it in every story.

angular/projects/spark-angular/src/lib/components/inputs/Search.stories.ts
```diff
@@ -72,6 +72,8 @@ export const searchInput = () => ({
           iconName="search"
           additionalClasses="sprk-b-InlineSearch__icon sprk-c-Icon--filled-current-color sprk-c-Icon--stroke-current-color"
           sprk-input-icon
+          ariaHidden="true"
+
```
afebbraro, line 76: Noticed an empty space here.

angular/projects/spark-angular/src/lib/components/inputs/Search.stories.ts
```diff
@@ -117,6 +119,8 @@ export const invalidSearchInput = () => ({
...
```
afebbraro, line 123: Noticed an empty space here.

html/base/inputs/Checkbox.stories.js
```diff
@@ -700,6 +702,8 @@ export const hugeInvalid = () => {
         <svg
           class="sprk-c-Icon sprk-b-ErrorIcon"
           viewBox="0 0 64 64"
+          aria-hidden="true"
```
afebbraro, line 705: Thank you for updating HTML as well!

afebbraro (approved):
Great! Just left one comment about extra spaces. Marking approved so you can merge after. Thank you!

PR #3974, opened 2021-04-27
Angular Autocomplete

bobwise:
## What does this PR do?
- Adds the `sprk-autocomplete` component to `spark-angular`.
- Variants for Error, Disabled, Huge, Huge Error, and Huge Disabled.
- Close the list by pressing `Escape` or clicking the background. Scroll the results with the arrow keys.
- `Enter` or clicking an item will emit the `itemSelectedEvent` if it is set on the component.
- `aria-controls` and `aria-owns` are generated by the component if needed.

Note that this PR *does not* include typing in the input to filter the list. We have a separate issue (2467748) to create the kitchen-sink examples [like we have for HTML](https://spark-testhtml.netlify.app/autocomplete.html).

Here is a deploy preview of the [kitchen sink examples](https://angry-morse-9dd8c4.netlify.app/).

image

### Associated Issue
Fixes 2256560

### Documentation
 - [x] Update Spark Docs

### Code
image

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [x] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)
  
### Accessibility Testing
  - [x] Axe browser extension
  - [x] Jaws Inspect

 - [x] Design Team
 - [x] Accessibility Team


angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete-result/sprk-autocomplete-result.directive.ts
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * @ignore
+   *
+   */
+  private _isHighlighted: boolean;
+
+  /**
+   * If true, this Autocomplete result will be highlighted.
+   */
+  @HostBinding('class.sprk-c-Autocomplete__result--active')
```
bobwise, line 62: This is the only `@HostBinding('class` in this PR. The other classes are added using `getClasses` functions.

styles/components/_autocomplete.scss
```diff
@@ -43,3 +43,8 @@
 .sprk-c-Autocomplete__result--active {
   background-color: $sprk-autocomplete-result-active-background-color;
 }
+
+/// Hide the Autocomplete Results
+.sprk-c-Autocomplete__results--hidden {
+  display: none;
+}
```
bobwise, line 50: New class so we don't have to use `sprk-u-Display--none`

styles/components/_autocomplete.scss
```diff
@@ -43,3 +43,8 @@
 .sprk-c-Autocomplete__result--active {
   background-color: $sprk-autocomplete-result-active-background-color;
 }
+
+/// Hide the Autocomplete Results
+.sprk-c-Autocomplete__results--hidden {
+  display: $sprk-autocomplete-results-hidden-display;
+}
```
bobwise, line 50: New class so we don't need to use `sprk-u-Display--none`

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.spec.ts
```diff
@@ -0,0 +1,929 @@
...
+  }));
+
+  let spy;
+  beforeEach(() => {
+    spy = jest.spyOn(console, 'warn').mockImplementationOnce(() => {});
+
+    fixture = TestBed.createComponent(WrappedAutocompleteComponent);
+    fixture.detectChanges();
+
+    component = fixture.debugElement.children[0].componentInstance;
```
bobwise, line 51: This allows us to call member functions like `component.hideResults()`.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
```diff
@@ -0,0 +1,425 @@
...
+
+  /**
+   * @ignore
+   * Copy the value of the id attribute on contentElement
+   * into the aria-controls attribute on triggerElement.
+   * Generate a unique ID if needed.
+   * @param triggerElement The element that will receive the aria-controls attribute.
+   * @param contentElement The element that will receive an id if needed.
+   */
+  generateAriaControls = (triggerElement, contentElement) => {
```
bobwise, line 346: `generateAriaControls` and `generateAriaOwns` are utility functions in HTML, but we've never needed them in Angular before. These can be refactored into utility functions if they're needed anywhere else.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete-results/sprk-autocomplete-results.directive.ts
```diff
@@ -0,0 +1,86 @@
+import {
+  Directive,
+  ElementRef,
+  OnInit,
+  HostListener,
```
RVMendoza, line 5: HostListener is not being used in this file at all, we can get rid of it.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete-result/sprk-autocomplete-result.directive.ts
RVMendoza: I see, since `aria-selected` and `--active` class are both changed upon, `isHighlighted`, you combined them.

I think that makes sense because they keep things closer together.
One thing I notice about `HostBinding` is that it adds the class to the custom element, not the actual HTML Element. But since this is a directive, this should be fine because it directly adds it to the element it's on.

https://codesandbox.io/s/elastic-river-7jw6f?file=/src/app/app.component.ts
When you click "activate" you see it only adds it to the custom element...which is not great.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.stories.ts
```diff
@@ -0,0 +1,599 @@
...
+    storyWrapper(
+      (storyContent) => `<div class="sprk-o-Box">${storyContent}<div>`,
+    ),
+  ],
+  parameters: {
+    info: `
+${markdownDocumentationLinkBuilder('autocomplete')}
+- Live demos of the Autocomplete are available on our
+[demo site](https://spark-testangular.netlify.app/autocomplete)
+([View Source](https://github.com/sparkdesignsystem/spark-starter-angular/tree/kitchen-sink/src/app/spark-docs/autocomplete-docs)).
```
bobwise, line 43: These URLs are currently dead links. These pages will be added as part of 2467748.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete-result/sprk-autocomplete-result.directive.ts
bobwise: Yes I think you're correct. Putting the class on the custom element is not a good idea because those custom elements are not valid HTML elements.

html/utilities/keypress.js
```diff
@@ -12,7 +12,7 @@ const isEnterPressed = (e) => e.key === 'Enter' || e.keyCode === 13;
 const isDownPressed = (e) => e.key === 'ArrowDown' || e.keyCode === 40;
 const isLeftPressed = (e) => e.key === 'ArrowLeft' || e.keyCode === 37;
 const isRightPressed = (e) => e.key === 'ArrowRight' || e.keyCode === 39;
-const isUpPressed = (e) => e.key === 'ArrowRight' || e.keyCode === 38;
+const isUpPressed = (e) => e.key === 'ArrowUp' || e.keyCode === 38;
```
bobwise, line 15: `key` is deprecated and all our supported browsers use `keyCode`, so I think this typo would only have caused problems in old browsers we don't support.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete-result/sprk-autocomplete-result.directive.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    fixture.detectChanges();
+  }));
+
+  it('should create itself', () => {
+    expect(component).toBeTruthy();
+  });
+
+  it('should add the correct base class', () => {
+    expect(element.classList.contains('sprk-c-Autocomplete__result')).toEqual(
```
afebbraro, line 51: Lets do a stricter check here by not checking if it contains the class but by checking all of the classes on the element to make sure they are only what we expect.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete-result/sprk-autocomplete-result.directive.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+      true,
+    );
+  });
+
+  it('should add classes if additionalClasses has a value', () => {
+    expect(element.classList.contains('testClass')).toEqual(true);
```
afebbraro, line 57: Lets do a stricter check here by not checking if it contains the class but by checking all of the classes on the element to make sure they are only what we expect.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete-result/sprk-autocomplete-result.directive.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+    directiveElement.clickedEvent.subscribe((g) => {
+      called = true;
+      done();
+    });
+    element.click();
+    expect(called).toEqual(true);
+  });
+
+  it('should add the class and aria attribute if isHighlighted is true', () => {
+    directiveElement.isHighlighted = true;
```
afebbraro, line 79: The naming of `isHighlighted` speaks to a specific style vs. a function. With styles being fluid over time, naming this something more specific to the expected functionality would ensure we don't have breaking changes in the future. Maybe something like `isSelected` could work.

afebbraro (changes requested):
hey, I only went through the first file so far so not done with review but wanted to send out these first few comments.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete-result/sprk-autocomplete-result.directive.ts
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Normally Spark would put this kind of behavior in ngOnChanges, but
+     * that lifecycle event is not triggered when an Input is updated directly
+     * from a parent component. This logic is moved into a setter on the Input
+     * so that it is triggered when sprk-autocomplete changes this value.
+     */
+    this._isSelected = value;
+
+    if (this._isSelected) {
+      var result = this.ref.nativeElement;
```
guern1kn, line 75: I don't see `result` getting set to a new value here, can update this to be `const`

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete-result/sprk-autocomplete-result.directive.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+    );
+  });
+
+  it('should add classes if additionalClasses has a value', () => {
+    expect(element.classList.toString()).toEqual(
+      'sprk-c-Autocomplete__result testClass',
+    );
+  });
+
+  it('should add a value for data-analytics if analyticsString has a value', () => {
```
guern1kn, line 64: Let's add the inverse of the data-analytics & data-id tests.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete-results/sprk-autocomplete-results.directive.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    );
+  });
+
+  it('should add classes if additionalClasses has a value', () => {
+    expect(element.classList.toString()).toEqual(
+      'sprk-c-Autocomplete__results testClass',
+    );
+  });
+
+  it('should add a value for data-analytics if analyticsString has a value', () => {
```
guern1kn, line 54: Inverse of these two tests.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete-results/sprk-autocomplete-results.directive.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    expect(component).toBeTruthy();
+  });
+
+  it('should add the correct base class', () => {
+    expect(emptyElement.classList.toString()).toEqual(
+      'sprk-c-Autocomplete__results',
+    );
+  });
+
+  it('should add classes if additionalClasses has a value', () => {
```
guern1kn, line 48: I like this way of having a strict check on the classList!

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
```diff
@@ -0,0 +1,421 @@
...
+        );
+      }
+    }
+
+    // if itemSelectedEvent is specified, also set that as the click event on each result
+    if (this.itemSelectedEvent) {
+      this.resultItems.forEach((element) => {
+        element.clickedEvent = this.itemSelectedEvent;
+      });
+    }
```
bobwise, line 300: This might need to move to another lifecycle hook. It looks like this is not being applied when the component is added/removed dynamically.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
```diff
@@ -0,0 +1,421 @@
...
+      }
+    }
+  }
+
+  /**
+   * Emit the itemSelectedEvent with the currently-highlighted Autocomplete
+   * result item.
+   */
+  selectHighlightedListItem(): void {
+    var selectedElement = this.resultItems.filter(
```
guern1kn, line 198: Update to `let`

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
```diff
@@ -0,0 +1,421 @@
...
+   * @param triggerElement The element that will receive the aria-controls attribute.
+   * @param contentElement The element that will receive an id if needed.
+   */
+  generateAriaControls = (triggerElement, contentElement) => {
+    const triggerAriaControls = triggerElement.getAttribute('aria-controls');
+    let contentId = contentElement.getAttribute('id');
+
+    // Warn if aria-controls exists but the id does not
+    if (triggerAriaControls && !contentId) {
+      console.warn(
```
guern1kn, line 356: I know we've talked about this a few times. My recollection is that we decided not to do warnings like this moving forward because we didn't want errors/warnings in the console of prod applications. I'd like to double check where we landed on that decision.

guern1kn (changes requested):
Great work on this, Bob! I have some questions about the console warnings before approving it.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
bobwise: I discussed this with @afebbraro in the team chat. The consensus so far is that Spark should not try to recreate functionality in the Axe browser extension. Spark should generate IDs and aria values and apply them when possible, but if the implementing engineer passes in invalid ids we're going to just leave them alone and trust Axe to flag the error.

If this sounds good to everyone, I'll change this PR so `generateAriaOwns` and `generateAriaControls` do not console.warn.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
```diff
@@ -0,0 +1,405 @@
...
+    // Do nothing if aria-controls and id both exist but don't match
+    if (contentId && triggerAriaControls && contentId !== triggerAriaControls) {
+      return;
+    }
+
+    // If we don't have a valid id, generate one with lodash
+    if (!contentId) {
+      contentId = uniqueId(`sprk_autocomplete_results_`);
+      contentElement.setAttribute('id', contentId);
+    }
```
bobwise, line 368: I think this can be simplified by just making that the default value on the `id` Input.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete-result/sprk-autocomplete-result.directive.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  /**
+   * @ignore
+   * Accepts a function to run when an item is clicked.
+   * This should be set automatically by sprk-autocomplete and should
+   * not be set manually by users.
+   */
+  @Output()
+  clickedEvent = new EventEmitter<any>();
```
bobwise, line 102: This does not have `readonly` because it is manually set by the parent component.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
```diff
@@ -0,0 +1,395 @@
...
+  }
+
+  /**
+   * @ignore
+   * This logic needs to happen in this event so that the correct DOM
+   * attributes will have been set.
+   */
+  ngAfterViewInit(): void {
+    if (this.input) {
```
afebbraro, line 295: We have two checks here for if there's an input and none of this code runs if there isn't one. What about if we added an early return here at the top instead. it would be like ` if (!this.input) { return; }`.  Then you could reduce the number of if statements and it would be less complex. It would read as "if there isn't an input then don't run any of this code."

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
```diff
@@ -0,0 +1,395 @@
...
+          'style',
+          'max-width:' + currentInputWidth + 'px',
+        );
+      }
+    }
+  }
+  /** @ignore */
+  isUpPressed = (e) => e.key === 'ArrowUp' || e.keyCode === 38;
+  /** @ignore */
+  isDownPressed = (e) => e.key === 'ArrowDown' || e.keyCode === 40;
```
afebbraro, line 337: These would great to add as utilities. We would use them in tabs and in modal. Could you add them as utilities? I made a utils folder in angular in my masthead pr with some I had made. We previously did not have a utils folder.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
```diff
@@ -0,0 +1,395 @@
...
+      return;
+    }
+
+    // Do nothing if aria-controls and id both exist but don't match
+    if (contentId && triggerAriaControls && contentId !== triggerAriaControls) {
+      return;
+    }
+
+    // set the value of aria-controls
+    triggerElement.setAttribute('aria-controls', contentId);
```
afebbraro, line 367: This should be updated to use Renderer and not interact with the DOM directly.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
```diff
@@ -0,0 +1,395 @@
...
+      return;
+    }
+
+    // Do nothing if aria-owns and id both exist but don't match
+    if (contentId && ownerAriaOwns && contentId !== ownerAriaOwns) {
+      return;
+    }
+
+    // set the value of aria-owns
+    ownerElement.setAttribute('aria-owns', contentId);
```
afebbraro, line 393: Should use renderer here too.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
```diff
@@ -0,0 +1,395 @@
...
+
+  /**
+   * @ignore
+   * Copy the value of the id attribute on contentElement
+   * into the aria-owns attribute on ownerElement.
+   * Generate a unique ID if needed.
+   * @param ownerElement The element that will receive the aria-owns attribute.
+   * @param contentElement The element that will receive an id if needed.
+   */
+  generateAriaOwns = (ownerElement, contentElement) => {
```
afebbraro, line 378: The params should be angular refs.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.module.ts
```diff
@@ -0,0 +1,36 @@
...
+    SprkAutocompleteResultsModule,
+    SprkAutocompleteResultModule,
+    SprkInputContainerModule,
+    SprkInputModule,
+    SprkFieldErrorModule,
+  ],
+  declarations: [SprkAutocompleteComponent],
+  exports: [
+    SprkAutocompleteComponent,
+    CommonModule,
```
afebbraro, line 26: We don't need to export CommonModule and RouterModule as those are done in their applications initial setup.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.stories.ts
```diff
@@ -0,0 +1,599 @@
...
+import { SprkAutocompleteComponent } from './sprk-autocomplete.component';
+import { SprkInputModule } from '../../directives/inputs/sprk-input/sprk-input.module';
+import { SprkInputContainerModule } from '../inputs/sprk-input-container/sprk-input-container.module';
+import { SprkIconModule } from '../sprk-icon/sprk-icon.module';
+import { SprkIconComponent } from '../sprk-icon/sprk-icon.component';
+import { SprkLabelModule } from '../../directives/inputs/sprk-label/sprk-label.module';
+import { SprkLabelDirective } from '../../directives/inputs/sprk-label/sprk-label.directive';
+import { SprkIconInputContainerModule } from '../inputs/sprk-icon-input-container/sprk-icon-input-container.module';
+import { SprkInputContainerComponent } from '../inputs/sprk-input-container/sprk-input-container.component';
+import { SprkFieldErrorModule } from '../../directives/inputs/sprk-field-error/sprk-field-error.module';
```
afebbraro, line 12: If all of the needed directives and components are importing and exported in the autocomplete module file then they all do not also need to imported here. This would mimic what an app would encounter. They should only need to import the autocomplete file. You will have to import the ones that are needed in the `subcomponents` property on the default story object so that we can display their documentation. But shouldnt have to import things like the `SprkInputModule`, `SprkIconComponent` etc. As those modules should be imported/exported in the autocomplete module as it uses them.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
bobwise: Yes definitely. I'll also move `generateAriaOwns` and `generateAriaControls` to utilities.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
bobwise: I changed a couple other spots as well.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.stories.ts
```diff
@@ -0,0 +1,599 @@
...
+  <sprk-autocomplete>
+    <sprk-input-container>
+      <div aria-live="polite" class="sprk-u-ScreenReaderText"></div>
+      <label
+        id="autocomplete-label2"
+        for="autocomplete-input2"
+        hasIcon="true"
+        sprkLabel
+      >
+        Search for a Fruit
```
afebbraro, line 188: Let's update this to keep the example on brand.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.stories.ts
```diff
@@ -0,0 +1,599 @@
...
+            sprk-c-Icon--stroke-current-color
+          "
+          sprk-input-icon
+          ariaHidden="true"
+        ></sprk-icon>
+        <input
+          id="autocomplete-input2"
+          name="autocomplete-input2"
+          type="text"
+          class="sprk-b-TextInput--has-svg-icon sprk-u-Width-100 sprk-b-TextInput--error"
```
afebbraro, line 210: Same thing with the sprk-u-Width-100 here.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.spec.ts
```diff
@@ -0,0 +1,915 @@
...
+      new KeyboardEvent('keydown', {
+        key: 'ArrowDown',
+      }),
+    );
+
+    fixture.detectChanges();
+    expect(component.advanceHighlightedItem).toBeCalledTimes(0);
+  });
+
+  it('should highlight the first item when Down is pressed and isOpen', () => {
```
afebbraro, line 390: Great job on all the test cases in here!

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.stories.ts
```diff
@@ -0,0 +1,577 @@
...
+            sprk-c-Icon--stroke-current-color
+          "
+          sprk-input-icon
+          ariaHidden="true"
+        ></sprk-icon>
+        <input
+          id="autocomplete-input3"
+          name="autocomplete-input3"
+          type="text"
+          class="sprk-b-TextInput--has-svg-icon sprk-u-Width-100"
```
afebbraro, line 294: `sprk-u-Width-100` could be removed with Katies Pr work here too.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.stories.ts
```diff
@@ -0,0 +1,577 @@
...
+          "
+          sprk-input-icon
+          ariaHidden="true"
+        ></sprk-icon>
+        <input
+          id="autocomplete-input4"
+          name="autocomplete-input4"
+          type="text"
+          placeholder="Search for a Fruit"
+          class="sprk-u-Width-100"
```
afebbraro, line 365: `sprk-u-Width-100` here too

html/utilities/keypress.js
afebbraro: Great catch!

src/pages/using-spark/components/autocomplete.mdx
```diff
@@ -63,6 +64,7 @@ help the user choose the correct data format for a task.
 <ComponentPreview
   componentName="autocomplete--default-story"
   hasHTML
+  hasAngular
```
afebbraro, line 67: Nice to see this!

styles/components/_autocomplete.scss
afebbraro: @bobwise  thanks for adding the new class! Lets update it to match the `is-hidden` naming convention though!

afebbraro (changes requested):
Left some comments! Great work! My next task will be checking the branch out and then viewing in browser. I don't expect to have any more comments but I will add them if anything comes up!

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
```diff
@@ -0,0 +1,398 @@
...
+   * @ignore
+   */
+  @HostListener('document:click', ['$event'])
+  onDocumentClick(event): void {
+    if (
+      this.ref.nativeElement &&
+      !this.ref.nativeElement.contains(event.target) &&
+      this.isOpen
+    ) {
+      this.hideResults();
```
RVMendoza, line 113: what happens when the if statement returns false? Nothing? do we need a `return`?

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.stories.ts
```diff
@@ -0,0 +1,577 @@
...
+  of the active item. Spark will apply this attribute automatically.
+- When a list item is highlighted (with the arrow keys), it should have
+  \`aria-selected="true"\`. Spark will apply this attribute automatically.
+- Each selectable list item should have \`role="option"\`.
+
+
+#### Requires Additional Engineering:
+In order to keep the Spark Autocomplete flexible enough to use in a wide
+  variety of use cases, some of the interactivity of the component is left
+  up to implementing teams.
```
RVMendoza, line 74: I like this explanation

RVMendoza (changes requested):

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete-result/sprk-autocomplete-result.directive.ts
bobwise: This `Output` has `ignore` because we don't want to document it, because if users try to set this value themselves weird things could happen. This comment is explaining why the `ignore` is here. I'll rewrite it to be more clear.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
bobwise: Nothing should happen when the if returns false. I don't think we need a `return`. The function returns on its own when it gets to the end of the function and we don't need to return any specific value to the caller.

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
```diff
@@ -0,0 +1,398 @@
...
+   * Track the index of the currently-highlighted Autocomplete result item.
+   */
+  highlightedIndex = -1;
+
+  /**
+   * @ignore
+   */
+  ngAfterContentInit(): void {
+    if (this.results) {
+      if (this.isOpen) {
```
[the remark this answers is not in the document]
RVMendoza: Talked in chat -- I understand where this is going now, it's more readable this way.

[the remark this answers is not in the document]
RVMendoza: had some chat conversations how we could omit the `else` here, but it was more of a code style thing -- I'll let @bobwise  decide what he likes the most!

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.stories.ts
```diff
@@ -0,0 +1,577 @@
...
+          "
+          sprk-input-icon
+          ariaHidden="true"
+        ></sprk-icon>
+        <input
+          id="autocomplete-input5"
+          name="autocomplete-input5"
+          type="text"
+          placeholder="Search for a Fruit"
+          class="sprk-u-Width-100 sprk-b-TextInput--error"
```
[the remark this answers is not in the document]
bobwise: The width 100 is part of the base input style now, so this is removed

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.component.ts
afebbraro: update, we are not moving these two to be utils at this time.

html/spark-exports.js
```diff
@@ -147,8 +147,6 @@ import {
 import { carousel } from './components/carousel';
 import { highlightBoard } from './components/highlight-board';
 import toggleAriaExpandedAttribute from './utilities/toggleAriaExpanded';
-import generateAriaControls from './utilities/generateAriaControls';
```
afebbraro, line 150: Did you mean to remove these?

bobwise: lol no

afebbraro (approved):
Looks good! Good work!

angular/projects/spark-angular/src/lib/components/sprk-autocomplete/sprk-autocomplete.stories.ts
```diff
@@ -125,7 +125,7 @@ export const defaultStory = () => ({
           id="autocomplete-input"
           name="autocomplete-input"
           type="text"
-          class="sprk-b-TextInput--has-svg-icon sprk-u-Width-100"
```
RVMendoza: Confirmed that this is removed because Katie's React Input PR will take care of adding width 100 to --has-svg class

RVMendoza (approved):
This is a really complicated component and you made this happen! Nice work!

guern1kn (approved):

PR #3979, opened 2021-05-06
Add property names to host bindings

jtneal:
## What does this PR do?
In Angular 10 (and TypeScript 3) `@HostBinding('class.example') true;` was interpreted such that `true` was the value and thus the HostBinding was applied properly. In Angular 11 (and TypeScript 4), `true` is instead interpreted as the property name, with no value, meaning a falsey value, and the HostBinding does not get applied. This change, backwards compatible with all previous versions, should fix these issues in newer versions of Angular and TypeScript.

### Associated Issue
n/a

### Code

 - [ ] Design Team
 - [ ] Accessibility Team


jtneal:
Tested in Angular 11

# Before

Screen Shot 2021-05-06 at 5 23 05 PM

# After

Screen Shot 2021-05-06 at 5 23 53 PM

PR #3990, opened 2021-05-27
Bug fix: Use type display one wide variable for type display one on wide viewports

afebbraro:
## What does this PR do
We had a display one wide var but we weren't using it for type one. I updated it to use it but need double check with design. Because the value we had for it is different than the current value. I updated it to be the old value so nothing about our type size will change. But I want to check if is that our type on wide viewports before was too small or it's that the wide var setting is too big.

here is the current settings
```
"one": {
          "comment": "The size of TypeDisplayOne when the viewport is narrower than $sprk-font-size-display-one-breakpoint.",
          "value": "2.75rem",
          "attributes": {
            "category": "size",
            "type": "font"
          },
          "file": "typography",
          "themable": true
        },
        "one-wide": {
          "comment": "The size of TypeDisplayOne when the viewport is wider than $sprk-font-size-display-one-breakpoint.",
          "value": "3.375rem",
          "file": "typography",
          "attributes": {
            "category": "size",
            "type": "font"
          },
          "themable": true
        },
```

I updated them to be same now, at 2.75rem (like we had it before).

### Associated Issue
Fixes 2494975

### Code
 - [x] Update var in type sass file

  - [x] Google Chrome
  - [ ] Google Chrome (Mobile)
  - [ ] Mozilla Firefox
  - [ ] Microsoft Edge
  - [ ] Apple Safari
  - [ ] Apple Safari (Mobile)
 
 - [ ] Design Team

### Screenshots
Add screenshots to help explain your PR if you'd like. However, this is not
expected.
This shows the same 2.75 value being applied.
Screen Shot 2021-05-26 at 7 58 25 PM


RVMendoza (approved):
Seems straightforward enough -- if design decided this then I'm satisfied code-wise!

PR #3995, opened 2021-06-01
Add alpha publish versions

afebbraro:


react/src/base/inputs/SprkInput/SprkInput.js
```diff
@@ -116,7 +116,7 @@ SprkInput.propTypes = {
    * A ref passed in will be attached to the
    * input element of the rendered component.
    */
-  forwardedRef: PropTypes.oneOf([PropTypes.shape(), PropTypes.func]),
+  forwardedRef: PropTypes.oneOfType([PropTypes.shape(), PropTypes.func]),
```
afebbraro, line 119: I was still getting errors when I rebased and ran storybook with react inputs and it was because we needed to make this change. We needed to use oneOfType and not oneOf for the proptypes. This is because oneOf is for literal values like 'cats, 'dogs' whereas oneOfType describes the type of values we can use like func`.

PR #4026, opened 2021-07-14
HTML Spinner Hotfix

bobwise:
## What does this PR do?
Adds a new data attribute to the HTML Spinner API that allows submit buttons to skip the `disabled` behavior.

image

### Documentation
 - [x] Update Spark Docs

### Code
 - [x] Build Component in HTML

image

  - [x] Google Chrome
  - [x] Google Chrome (Mobile)
  - [x] Mozilla Firefox
  - [ ] Microsoft Edge
  - [x] Apple Safari
  - [x] Apple Safari (Mobile)
  
### Accessibility Testing
  - [x] Axe browser extension
  - [ ] Jaws Inspect

 - [ ] Design Team
 - [ ] Accessibility Team

### DIFF
--- a/html/components/spinners.js
+++ b/html/components/spinners.js
@@ -32,13 +32,19 @@ const setSpinning = (element, options) => {
   const spinningAriaLabel = options.ariaLabel || 'Loading';
   const ariaValueText = options.ariaValueText || 'Loading';
   const role = options.role || 'progressbar';
+  const doNotDisable = options.doNotDisable || false;
 
   el.classList.add('sprk-c-Button--has-spinner');
   el.setAttribute('aria-label', spinningAriaLabel);
   el.setAttribute('data-sprk-spinner-text', el.textContent);
-  el.setAttribute('disabled', '');
   el.setAttribute('data-sprk-has-spinner', 'true');
   el.setAttribute('style', `width: ${width}px`);
+
+  // This flag should be used for submit buttons so that
+  // the disabled attribute does not suppress the submit behavior.
+  if (!doNotDisable) {
+    el.setAttribute('disabled', '');
+  }
 
   el.innerHTML = `
     <div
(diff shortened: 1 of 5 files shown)


guern1kn (approved):

html/components/spinners.js
```diff
@@ -84,6 +90,37 @@ const spinners = () => {
       }
     });
   });
+
+  /**
+   * This selector is exactly the same as data-sprk-spinner="click" except
+   * the disabled attribute will not be added to the button when it is
+   * spinning.
+   */
+  getElements('[data-sprk-spinner="is-not-disabled"]', (spinnerContainer) => {
```
RVMendoza, line 99: is this implying that it either has `click` or `is-not-disabled`? 

What are you thoughts on if we grabbed elements with [data-sprk-spinner] and then do an in statement based on if its value is `click` or `is-not-disabled`so we're repeating less code?

RVMendoza: Perhaps something like this: https://stackoverflow.com/questions/7084557/select-all-elements-with-a-data-xxx-attribute-without-using-jquery

and then using `getAttribute` to switch between the two

html/components/spinners.js
```diff
@@ -32,14 +32,20 @@ const setSpinning = (element, options) => {
...
   el.classList.add('sprk-c-Button--has-spinner');
   el.setAttribute('aria-label', spinningAriaLabel);
   el.setAttribute('data-sprk-spinner-text', el.textContent);
-  el.setAttribute('disabled', '');
   el.setAttribute('data-sprk-has-spinner', 'true');
   el.setAttribute('style', `width: ${width}px`);
 
+  // This flag should be used for submit buttons so that
+  // the disabled attribute does not suppress the submit behavior.
+  if (!doNotDisable) {
```
RVMendoza, line 45: this double negative is throwing me off, perhaps something more brain friendly like `hasDoNotDisable`

RVMendoza: it's only *slightly* less confusing though so if you have a better idea, go for it haha

RVMendoza (approved):
these changes are great! I'm glad we could find a more efficient way of doing the same thing!

PR #4039, opened 2021-07-21
Flip accordion arrow icon & add additional class for optional top border

rachelsuter:
## What does this PR do?
Flip accordion arrow icon to be facing down when accordion is closed and up when open. Add additional accordion class for optional top border.

### Associated Issue
2474675

### DIFF
--- a/angular/projects/spark-angular/src/lib/components/sprk-accordion/sprk-accordion.stories.ts
+++ b/angular/projects/spark-angular/src/lib/components/sprk-accordion/sprk-accordion.stories.ts
@@ -52,7 +52,7 @@ the [Toggle Component](/docs/components-toggle--default-story) instead.
 
 export const defaultStory = () => ({
   template: `
-    <sprk-accordion>
+    <sprk-accordion additionalClasses="sprk-c-Accordion--no-top-border">
       <sprk-accordion-item
         heading="This is an accordion heading"
         additionalClasses="sprk-u-mbs"
--- a/angular/projects/spark-angular/src/lib/components/sprk-accordion/sprk-accordion.stories.ts
+++ b/angular/projects/spark-angular/src/lib/components/sprk-accordion/sprk-accordion.stories.ts
@@ -55,7 +55,6 @@ export const defaultStory = () => ({
     <sprk-accordion additionalClasses="sprk-c-Accordion--no-top-border">
       <sprk-accordion-item
         heading="This is an accordion heading"
-        additionalClasses="sprk-u-mbs"
         idString="accordion-item-1"
         analyticsString="object.action.event"
       >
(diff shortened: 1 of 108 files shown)


styles/components/_accordions.scss
```diff
@@ -11,6 +11,10 @@
   width: 100%;
 }
 
+.sprk-c-Accordion--no-top-border {
+  border-top: none;
```
guern1kn, line 15: Let's make this a variable.

react/src/components/accordions/SprkAccordion.stories.js
```diff
@@ -19,7 +19,7 @@ the [Toggle Component](/docs/components-toggle--default-story) instead.
 };
 
 export const defaultStory = () => (
-  <SprkAccordion>
+  <SprkAccordion additionalClasses="sprk-c-Accordion--no-top-border">
```
afebbraro, line 22: The issue has it so that this is an option. So we would want to not add it here as the default.

html/components/accordion.stories.js
```diff
@@ -32,7 +32,9 @@ export const defaultStory = () => {
   }, []);
 
   return `
-<ul class="sprk-c-Accordion sprk-o-VerticalList">
+<ul 
+  class="sprk-c-Accordion sprk-c-Accordion--no-top-border sprk-o-VerticalList"
```
afebbraro, line 36: This is optional so we don't want to add it as the default

angular/projects/spark-angular/src/lib/components/sprk-accordion/sprk-accordion.stories.ts
```diff
@@ -52,7 +52,7 @@ the [Toggle Component](/docs/components-toggle--default-story) instead.
 
 export const defaultStory = () => ({
   template: `
-    <sprk-accordion>
+    <sprk-accordion additionalClasses="sprk-c-Accordion--no-top-border">
```
afebbraro, line 55: This is optional so we don't want to add it as the default

react/src/components/accordions/SprkAccordion.stories.js
afebbraro: We would usually make the component have a new prop (for react) and a new Input (for angular) that would add this class instead of additionalClasses. If that is something that you could update this PR with, that would be better than having this as an additionalClass.

afebbraro (changes requested):
Hey Rachel! Thank you for doing this issue! I added a couple comments on things. If you have any questions reach out to any of us.

RVMendoza (approved):
Nice work! Once Amber's stuff is addressed, it's good to me!

styles/components/_accordions.scss
```diff
@@ -11,6 +11,10 @@
   width: 100%;
 }
 
+.sprk-c-Accordion--no-top-border {
```
guern1kn, line 14: Add a comment above this to explain what it's doing. See the class below for reference. This will also make sure it gets added to the modifiers documentation for Accordion.

PR #4059, opened 2021-08-13
add terser

guern1kn:
## What does this PR do?
Add terser and update rollup babel plugin


afebbraro:
nice, thank you! We'll need to verify the `dist`output. So i'll find some time on Monday to check the branch down and make sure the output is the same with the new plugins. It'll be a full package test too. But, no code changes will be needed to this PR!

afebbraro:
Checked out the branch and saw that some react code is being bundled in the output and so we are closing for now and will re-open once we decide next steps. Thanks again!

PR #4060, opened 2021-08-19
Component status page

rachelsuter:
## What does this PR do?
Build component landing and status page.

### Associated Issue
2571011

### DIFF
--- a/src/components/Status.js
+++ b/src/components/Status.js
@@ -1,0 +1,141 @@
+import React from 'react';
+import { SprkTable } from '@sparkdesignsystem/spark-react';
+import {
+  SprkLink,
+  SprkIcon,
+  SprkDivider,
+  SprkToggle,
+  SprkAccordion,
+  SprkAccordionItem
+} from '@sparkdesignsystem/spark-react';
+
+const Status = ({
+    statusName,
+    href
+}) => {
+    if (statusName == "ready") {
+        return (
+            <div>
+                <SprkIcon className="
+                    docs-status__Ready 
+                    sprk-c-Icon--filled
+                    sprk-c-Icon--l"
+                    iconName="check-mark-filled"
+                />
+                <SprkLink
+                    element="a"
+                    href={href}
+                    variant="simple"
+                    additionalClasses="sprk-u-Color--green"
+                >
+                    Ready
+                </SprkLink> 
+            </div>
+        );
+    } else if (statusName == "underReview") {
+        return (
+            <div>
+                <SprkIcon className="
+                    docs-status__UnderReview
+                    sprk-c-Icon--filled
+                    sprk-c-Icon--l"
+                    iconName="caution-filled"
+                />
+                <SprkLink
+                    element="a"
+                    href={href}
+                    variant="simple"
+                    additionalClasses="sprk-u-Color--yellow"
+                >
+                    Under Review
+                </SprkLink>
+            </div>
+        );
+    } else if (statusName == "updated") {
+        return (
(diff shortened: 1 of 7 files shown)


src/scss/components/_status.scss
```diff
@@ -0,0 +1,122 @@
...
+// Status links hover colors
+.docs-link__green:hover {
+    border-bottom: 2px solid $sprk-green;
+}
+.docs-link__yellow:hover {
+    border-bottom: 2px solid #B46417;
+}
+.docs-link__blue:hover {
+    border-bottom: 2px solid $sprk-blue;
+}
```
bobwise, line 60: We also need `:focus` styles for keyboard use.

src/components/Status.js
```diff
@@ -0,0 +1,146 @@
...
+}) => {
+    if (statusName === "ready") {
+        return (
+            <div>
+                <SprkIcon className="
+                    docs-status__Ready 
+                    sprk-c-Icon--filled
+                    sprk-c-Icon--l"
+                    iconName="check-mark-filled"
+                />
```
bobwise, line 21: The icon for each of these statuses should have `aria-hidden="true"` so they are hidden from screen readers and other assistive technology. We don't need to provide a label for these images (like with `aria-label`) because the link text immediately following this image adequately describes the link.

bobwise (changes requested):

src/components/Status.js
```diff
@@ -0,0 +1,146 @@
...
+                    docs-status__UnderReview
+                    sprk-c-Icon--filled
+                    sprk-c-Icon--l"
+                    iconName="caution-filled"
+                />
+                <SprkLink
+                    element="a"
+                    href={href}
+                    variant="simple"
+                    additionalClasses="docs-status-text__Orange docs-link__yellow"
```
RVMendoza, line 45: something feels weird about why these have two colors. Is it orange or is it yellow? I'm noticing that `docs-link__yellow` has an orange underline. Why is that and why are they in separate classes?

src/scss/components/_status.scss
```diff
@@ -0,0 +1,122 @@
...
+
+
+// Status additional info question mark styles
+.docs-status-question-mark {
+    height: $sprk-icon-m;
+    width: $sprk-icon-m;
+    stroke-width: $sprk-icon-stroke-width;
+    margin-left: 6px;
+}
+.docs-status-question-mark__blue {
```
RVMendoza, line 70: We're on the right track, I would do some renaming all these classes to focus more on the status than the color. You were doing this with the Status classes. Combine some classes too, it seems like you're separately declaring color and underline color in different classes.

# Reasons
- Colors could change one day
- They're more descriptive about what will change in the component
- Things are more consistent in the code, making it easier to read.

RVMendoza (changes requested):

src/components/Status.js
```diff
@@ -0,0 +1,146 @@
...
+  SprkIcon,
+  SprkTooltip
+} from '@sparkdesignsystem/spark-react';
+
+const Status = ({
+    statusName,
+    href,
+    additionalInfo
+}) => {
+    if (statusName === "ready") {
```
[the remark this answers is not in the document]
RVMendoza: function name possibilities:
- getIconName
- getStatusText

bobwise (changes requested):

bobwise:
Great job @rachelsuter!!

yodasw16:
Closing this PR until a later date.

PR #4077, opened 2021-09-28
Footer inline link fix

rachelsuter:
## What does this PR do?
This PR fixes the footer inline link bug that appears on the visited state. The underline color was transparent on visited and should now be white. Changed the tokens style values for visited-border-bottom. Should apply to all frameworks.

### Associated Issue
2689258


RVMendoza (changes requested):
Nice! I liked that you saw the existing values and used those.

One final check: I'm pretty sure there's another file that's generated from this change. 

For example: https://github.com/sparkdesignsystem/spark-design-system/pull/4067/files

This generated changes for an android file AND a swift file.

Try building your project once more and see if there's any new files generated. For one final check, go to our HTML folder and run `npm run build`.

If everything's good, and nothing else is generated, then this is good for me!

RVMendoza:
image
confirmed that this PR fixes the visual change on :visited

afebbraro (approved):

RVMendoza (approved):
Built myself, did not see new generated style files.

